PR #115, opened 2023-07-12
Regenerate Java SDK using recent OpenAPI Specification

mgyucht:
## Changes
Spec commit sha: 17a3f7fe6 (7 July 2023)

Breaking Changes:
* Use CONSTANT_CASE for Enum constants. Many enums already use constant case in their definition, but some constants (like the SCIM Patch schema name) includes symbols `:` and numbers, so the SDK cannot use the enum value as the name. To do this, the `@JsonProperty` annotation is used to define the serialized form of the enum.
* Replace Query type with AlertQuery in sql.Alert class.
* Removal of User.isDbAdmin and User.profileImageUrl.

Changes:
* Introduce CleanRooms API
* Fix System Tables update HTTP method
* Introduce TablesAPI.update()
* Introduce Group.meta property
* Fix SCIM Patch implementation
* Introduce BaseRun.jobParameters and BaseRun.triggerInfo
* Introduce CreateJob.parameters
* Fix spelling in file arrival trigger configuration
* Introduce GitSource.jobSource
* Introduce RepairRun.rerunDependentTasks
* Introduce Resolved*Values classes, RunIf, and RunJobTask
* Introduce TaskNotificationSettings

Later follow-up:
* Names should split on Pascal-case word boundaries (see CloudProviderNodeStatus). This is an OpenAPI code gen change that needs to be made.


pietern:
Rest LGTM.

pietern (approved):

PR #116, opened 2023-07-17
Updated CLI token source `parseExpiry` method to account for different time format

tanmay-db:
## Changes

Add functionality to check for different date format for expiry parsing in CLI token source. Also add pattern matching for flexibility between different java versions

## Tests

Added unit tests, since `LocalDateTime` support upto nano second precision -- tested for less and more digits as well.

### DIFF
--- a/databricks-sdk-java/src/test/java/com/databricks/sdk/core/CliTokenSourceTest.java
+++ b/databricks-sdk-java/src/test/java/com/databricks/sdk/core/CliTokenSourceTest.java
@@ -29,7 +29,7 @@ public class CliTokenSourceTest {
   @Test
   public void testParseExpiryWithMoreThanNineNanoSecondDigits() {
     try {
-      LocalDateTime parsedDateTime = CliTokenSource.parseExpiry("2023-07-17T09:02:22.33061221987Z");
+      CliTokenSource.parseExpiry("2023-07-17T09:02:22.33061221987Z");
     } catch (DateTimeParseException e) {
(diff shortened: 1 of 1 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/CliTokenSource.java
```diff
@@ -37,10 +37,13 @@ public CliTokenSource(
     this.getAllEnv = getAllEnv;
   }
 
-  private static LocalDateTime parseExpiry(String expiry) {
+  static LocalDateTime parseExpiry(String expiry) {
```
tanmay-db, line 40: Making package private

cdkrot (approved):
This looks good to me. May be we can also add test into cli to assert those are the only formats it can output?

tanmay-db:
> This looks good to me. May be we can also add test into cli to assert those are the only formats it can output?

Not sure if we can add test into az cli since we don't build it. Probably in the databricks cli we can (not sure if there already is) but that would be external of Java SDK.

cdkrot:
Into *databricks cli I mean

The problem I have reported on slack was with our cli, not az cli (but there are a LOT problems with az cli as well)

tanmay-db:
> Into *databricks cli I mean
> 
> The problem I have reported on slack was with our cli, not az cli (but there are a LOT problems with az cli as well)
cc: @pietern for visibility on CLI

nfx (approved):

PR #125, opened 2023-07-25
Handle nested query parameters in ApiClient.withQuery

mgyucht:
## Changes
The Query History list API filter_by query parameter is modeled by a dictionary, rather than a primitive type, defining the allowed filters including query_start_time_range, statuses, user_ids, and warehouse_ids. To be compatible with gRPC transcoding, query parameters modeled by message types as opposed to primitives need to be separated into one query parameter per nested field, where the key is the path to that query parameter. For example:
```
new ListQueryHistoryRequest()
  .setFilterBy(new QueryFilter().setUserIds(Arrays.asList(123L, 456L)))
```
becomes
```
filter_by.user_ids=123&filter_by.user_ids=456
```
For this to be compatible with the requests library we use today, we need to recursively compute the path of each field in the request object that is annotated with the `QueryParam` annotation, then serialize the value according to Jackson.

As part of this, I've also generalized the `Request` class to support repeated query parameter values for lists (see the added integration test).

This resolves one of the problems from https://github.com/databricks/databricks-sdk-py/issues/99. The issue with the conflict between filter_by and next_page_token will be resolved by the backend service.

## Tests
Added an integration test for SQL Query History API.


databricks-sdk-java/src/main/java/com/databricks/sdk/core/GrpcTranscodingQueryParamsSerializer.java
```diff
@@ -0,0 +1,136 @@
...
+          Short.class,
+          int.class,
+          Integer.class,
+          long.class,
+          Long.class,
+          float.class,
+          Float.class,
+          double.class,
+          Double.class,
+          String.class);
```
tanmay-db, line 92: We can probably reuse this in future, it might be good to have this in a separate file.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/http/Request.java
```diff
@@ -40,7 +40,12 @@ public Request withHeader(String key, String value) {
   }
 
   public Request withQueryParam(String key, String value) {
-    query.put(key, value);
+    List<String> values = query.get(key);
+    if (values == null) {
+      values = new ArrayList<>();
+    }
+    values.add(value);
+    query.put(key, values);
```
tanmay-db, line 48: Should we handle the use case for when key exists?

databricks-sdk-java/src/main/java/com/databricks/sdk/support/Paginator.java
```diff
@@ -71,6 +71,9 @@ private Iterator<T> outerIterator() {
     return new Iterator<T>() {
       @Override
       public boolean hasNext() {
+        if (currentPage == null) {
+          return false;
+        }
```
tanmay-db, line 76: Curious -- why wasn't this required before?

tanmay-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/support/Paginator.java
mgyucht: This is kind of a separate bugfix, but I think that the query history API does not include the "results" field at all if there are no results. This is handled in flipNextPage line 63, but not in hasNext().

databricks-sdk-java/src/main/java/com/databricks/sdk/core/http/Request.java
mgyucht: if the key exists, values is not null, so we just skip the if block. Note that I've refactored this, as the put at the end is only needed in the case that values is null.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GrpcTranscodingQueryParamsSerializer.java
mgyucht: If we need to reuse it, we can make it public at some point. It's hard to figure out what class/file this would exist in that would be useful in other contexts, so I would rather leave it as is until we have two or three other things that it makes sense to group this with.

PR #135, opened 2023-08-14
Propagate Request Headers through API Client

mgyucht:
## Changes
Based on changes in https://github.com/databricks/databricks-sdk-go/pull/572.

This PR adds initial support for headers in ApiClient.java. The `*Impl` classes, when making a request, will construct a map of headers to be included in the request. These are passed through a new parameter in each of GET/PUT/POST/PATCH/DELETE.

In the future, to support dynamic headers, we can modify our codegen template to add those fields to the headers map before calling the ApiClient.

Alternatives considered:
* **Adding headers in request objects directly**: this would not require any change in the ApiClient interface. Instead, we would introduce a new annotation for headers, much like what we do for query parameters, and reflectively scan the fields of each request object. We could do this, but for headers that have a fixed value (like `Content-Type` and `Accept`), these fields would essentially be private final fields that we could iterate through. Users can never interact with these fields, so it seems a bit unusual to have them in the request structure, a public interface.


pietern (approved):

.codegen/impl.java.tmpl
```diff
@@ -26,18 +27,26 @@ class {{.PascalName}}Impl implements {{.PascalName}}Service {
     String path = {{if .PathParts -}}
       String.format("{{range  .PathParts}}{{.Prefix}}{{if or .Field .IsAccountId}}%s{{end}}{{ end }}"{{ range .PathParts }}{{if .Field}}, request.get{{.Field.PascalName}}(){{ else if .IsAccountId }}, apiClient.configuredAccountID(){{end}}{{ end }})
     {{- else}}"{{.Path}}"{{end}};
+    {{ template "headers" . }}
     {{if .Response -}}
-      {{- if .Response.ArrayValue -}} return apiClient.getCollection(path, null, {{template "type" .Response.ArrayValue}}.class);
-      {{- else if .Response.MapValue -}} return apiClient.getStringMap(path, request);
+      {{- if .Response.ArrayValue -}} return apiClient.getCollection(path, null, {{template "type" .Response.ArrayValue}}.class, headers);
```
nfx, line 32: the diff seems to be larger than it should be. did you try solving it with method overloading?
```suggestion
      {{- if .Response.ArrayValue -}} return apiClient.getCollection(path, null, {{template "type" .Response.ArrayValue}}.class{{if .HasRequestHeaders}}, headers{{end}});
```

nfx (dismissed):

mgyucht (commented):
Some old comment?

PR #136, opened 2023-08-17
Set necessary headers when authenticating via Azure CLI

mgyucht:
## Changes
The Java SDK request authentication logic is inconsistent between the Azure login types: for service principal auth, the SDK correctly adds the X-Databricks-Azure-Workspace-Resource-Id when configured, but this is missed for Azure CLI auth. Additionally, when logging in via Azure CLI using a service principal, the service management token must also be fetched from the CLI.

This PR fixes this by defining the logic to attach these header in a common function that is used by all Azure-specific authentication types.

See https://github.com/databricks/databricks-sdk-go/pull/584 for the same change in the Go SDK.
See https://github.com/databricks/databricks-sdk-py/pull/290 for the same changes in the Python SDK.
## Tests
- [x] Unit tests to cover the two scenarios for Azure CLI w.r.t. management endpoint token fetching, and one to verify that X-Databricks-Azure-Workspace-Resource-Id is included when using Azure CLI.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/AzureCliCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/AzureCliCredentialsProvider.java
@@ -43,6 +43,7 @@ public class AzureCliCredentialsProvider implements CredentialsProvider, AzureUt
       try {
         mgmtTokenSource.getToken();
       } catch (Exception e) {
+        LOG.debug("Not including service management token in headers", e);
         mgmtTokenSource = null;
       }
       CliTokenSource finalMgmtTokenSource = mgmtTokenSource;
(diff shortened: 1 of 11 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/AzureCliCredentialsProvider.java
```diff
@@ -37,12 +37,24 @@ public HeaderFactory configure(DatabricksConfig config) {
...
+      CliTokenSource mgmtTokenSource =
+          tokenSourceFor(config, config.getAzureEnvironment().getServiceManagementEndpoint());
       tokenSource.getToken(); // We need this for checking if Azure CLI is installed.
+      try {
+        mgmtTokenSource.getToken();
+      } catch (Exception e) {
+        LOG.debug("Not including service management token in headers", e);
+        mgmtTokenSource = null;
+      }
+      CliTokenSource finalMgmtTokenSource = mgmtTokenSource;
```
pietern, line 49: What is final about it? Seems like the alias is not necessary below.

pietern (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/AzureCliCredentialsProvider.java
mgyucht: Java style doesn't allow capturing non-final variables in closures. https://www.baeldung.com/java-lambda-effectively-final-local-variables#capturing-lambdas

PR #138, opened 2023-08-18
Support inputstream for request and response bodies

mgyucht:
## Changes
This PR adds support for non-application/json requests and responses. This PR is divided into two parts:

**Support setting headers per request.** The main methods of the ApiClient interface are expanded to accept one new parameter,`Map<String, String> headers`, and all callers are expected to pass a map of headers to be included in the request. As the allowed content types for requests and responses are known from the OpenAPI specification, impls must construct the request header map in code generation and pass it to the ApiClient. The default "Content-Type: application/json" header is removed, as each request should specify its own Content-Type and Accept headers. (This is merged in https://github.com/databricks/databricks-sdk-java/pull/135.)

**Add support for streaming request and response bodies to support non-application/json requests/responses.** Today, serialization of requests is done in the ApiClient. This implies that ApiClient needs to be able to serialize a request purely based on the parameters provided to it via headers, the request body, etc. In this proposal, the serialization is coupled to the request type: InputStream indicates that the client has already serialized the request, so it can be provided directly to the underlying HttpClient library, and any other type will be passed through Jackson's ObjectMapper for serialization to JSON. Internally, request and response bodies are modeled as InputStreams rather than byte arrays to support streaming requests and responses, though for non-streaming requests and responses, a string value for the body is captured in the `debugBody` field of Request and Response, respectively.

One important caveat about streaming responses is that callers are required to close the streams when they are done using them. This releases the underlying HTTP connection back to the connection pool for use in subsequent requests. Otherwise, customers could encounter deadlocks waiting for connections. For this, we recommend that users use the try-with-resources model as demonstrated in integration test FilesIT.java.

## Tests
- [x] Integration test for Files IT exercises uploading and downloading files using the streaming pathways.
- [x] Existing integration tests cover the pre-existing non-streaming pathways.


databricks-sdk-java/src/main/java/com/databricks/sdk/core/http/Request.java
```diff
@@ -140,7 +156,7 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(method, url, query, body);
+    return Objects.hash(method, url, query, debugBody);
```
mgyucht, line 159: leave comment for testing only

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GrpcTranscodingQueryParamsSerializer.java
```diff
@@ -116,7 +117,7 @@ private static Map<String, Object> flattenObject(Object o) {
           continue;
         }
         // check if object is a primitive type or a collection of some kind
-        if (primitiveTypes.contains(f.getType()) || Iterable.class.isAssignableFrom(f.getType())) {
+        if (primitiveTypes.contains(f.getType()) || Iterable.class.isAssignableFrom(f.getType()) || f.getType().isEnum()) {
```
mgyucht, line 120: Move to another PR

databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
```diff
@@ -53,22 +56,54 @@ private CloseableHttpClient makeClosableHttpClient() {
...
+      return new Response(in, statusLine.getStatusCode(), statusLine.getReasonPhrase(), hs);
+    }
+
+    // The Databricks SDK is currently designed to treat all non-application/json responses as
+    // InputStreams, leaving the caller to decide how to read and parse the response. The caller
+    // is responsible for closing the InputStream to release the HTTP Connection.
+    //
+    // If the API call fails, the server will respond with an application/json response, which
+    // the client reads fully and closes, even if the request Accept header is not application/json.
+    // The error is then thrown as an exception in ApiClient.
```
mgyucht, line 83: This needs to be implemented properly

mgyucht: Change to `if expected not json && got not json <=> stream`

databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/ApiErrors.java
```diff
@@ -71,15 +77,24 @@ private static ApiErrorBody parseUnknownError(Response response, IOException err
...
           String.format(
-              "Response from server (%s) %s: %s",
-              response.getStatus(), response.getBody(), err.getMessage()));
+              "Response from server (%s) %s: %s", response.getStatus(), body, err.getMessage()));
     }
     return errorBody;
   }
+
+  private static String convert(InputStream in) {
+    try (BufferedReader br =
```
mgyucht, line 93: Put into ByteArrayOutputStream

databricks-sdk-java/src/main/java/com/databricks/sdk/service/files/FilesAPI.java
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Get the status of a file or directory.
+   *
+   * <p>Returns the status of a file or directory.
+   */
+  public FileInfo getStatus(GetStatusRequest request) {
+    return impl.getStatus(request);
+  }
+
+  public void uploadFile(InputStream contents, String filePath) {
```
mgyucht, line 69: Move body parameter to the end

mgyucht:
Need to split into two PRs.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/BodyLogger.java
```diff
@@ -48,8 +48,11 @@ public String redactedDump(String body) {
       Object result = recursiveMarshal(rootNode, maxBytes);
       return mapper.writerWithDefaultPrettyPrinter().writeValueAsString(result);
     } catch (JsonProcessingException e) {
-      // Unable to unmarshal means the body isn't JSON (or something else)
-      return String.format("[unable to marshal: %s]", e.getMessage());
+      // Unable to unmarshal means the body isn't JSON
+      if (body.length() > maxBytes) {
+        return body.substring(0, maxBytes - 3) + "...";
```
pietern, line 53: This value should be escaped, seeing as it ends up in log files. If we log a files API response it can be any type of binary data and it can mess up your terminal.

pietern (commented):
Approach of making API client be streaming aware LGTM.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/BodyLogger.java
mgyucht: For now, I think this is safe-ish. The main value that triggers this case is "<InputStream>", since that is what we set debugBody to when streaming a response. Otherwise, it behaves the same as existing logging. I'll change this impl to reuse onlyNBytes, and we can pick this up as a follow-up taks.

mgyucht: Also for future reference: the implementation currently splits the debug body string by newlines and prepend the `> ` prefix, so it should not be possible for attackers to trigger printing out specific log files that don't have that prefix.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
```diff
@@ -53,22 +56,56 @@ private CloseableHttpClient makeClosableHttpClient() {
...
+                } catch (Exception e) {
+                  throw new DatabricksException("Unable to close connection", e);
+                }
+              });
+      return new Response(
+          in, statusLine.getStatusCode(), statusLine.getReasonPhrase(), hs, inputStream);
+    }
+
+    try (InputStream inputStream = entity.getContent()) {
+      String body = IOUtils.toString(inputStream, Charset.defaultCharset());
```
mgyucht, line 105: UTF-8

databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/ApiErrors.java
```diff
@@ -71,15 +73,31 @@ private static ApiErrorBody parseUnknownError(Response response, IOException err
       errorBody.setErrorCode(errorCode.replaceAll(" ", "_").toUpperCase());
     }
 
-    Matcher messageMatcher = HTML_ERROR_REGEX.matcher(response.getBody());
+    String body = convert(response.getBody());
```
mgyucht, line 76: Comment that error responses are fully consumed so don't need to be closed.

tanmay-db (approved):
Went over the PR with Miles, LGTM.

PR #139, opened 2023-08-18
Update OpenAPI Spec & Merge Files API

mgyucht:
## Changes
This PR adds recent changes from OpenAPI to the Java SDK. Notably, it adds support for the Files API. The integration test for this API can only be run in UC workspaces, so this PR also adds support for UC-only integration tests at the workspace- and account-level.

## Tests
- [x] New integration test for Files API to verify that it works as expected.


.codegen/impl.java.tmpl
```diff
@@ -26,18 +28,36 @@ class {{.PascalName}}Impl implements {{.PascalName}}Service {
...
+apiClient.{{.Verb}}(path
+  {{- if .Request}}, {{ template "request-param" .}}{{end}}
+  , {{ if not .Response -}}Void
+  {{- else if .IsResponseByteStream}}InputStream
+  {{- else}}{{template "type" .Response}}{{- end -}}.class
+  , headers);
+{{- end }}
+
+{{ define "headers" -}}
+    Map<String, String> headers = new HashMap<>();
```
nfx, line 55: this results in the unnecessary diff and headers variable for literally every single method. surround it with `if` and make sure there's no diff in the unaffected methods
```suggestion
    {{if .FixedRequestHeaders}}
    ... what you have here
    {{end}}
```

.codegen/impl.java.tmpl
```diff
@@ -26,18 +28,36 @@ class {{.PascalName}}Impl implements {{.PascalName}}Service {
...
-      {{- else -}} return apiClient.{{.Verb}}(path{{if .Request}}, request{{end}}{{if .Response -}}, {{if .Response.ArrayValue }}Collection
-                        {{- else if .Response.MapValue }}Map
-                        {{- else}}{{template "type" .Response}}
-                        {{- end -}}{{else}}, Void{{end}}.class);{{end}}
-    {{- else -}}apiClient.{{.Verb}}(path{{if .Request}}, request{{end}}{{if .Response -}}, {{if .Response.ArrayValue }}Collection
-      {{- else if .Response.MapValue }}Map
-      {{- else}}{{template "type" .Response}}
-      {{- end -}}{{else}}, Void{{end}}.class);
-    {{end}}
+    {{ template "headers" . -}}
```
nfx, line 31:
```suggestion
    {{- template "headers" . -}}
```

to cut whitespace before

.codegen/impl.java.tmpl
```diff
@@ -26,18 +28,36 @@ class {{.PascalName}}Impl implements {{.PascalName}}Service {
...
+    {{ if not .Response -}}
+    {{ template "api-call" . }}
+    {{- else if .Response.ArrayValue -}} return apiClient.getCollection(path, null, {{template "type" .Response.ArrayValue}}.class, headers);
+    {{- else if .Response.MapValue -}} return apiClient.getStringMap(path, {{ template "request-param" .}}, headers);
+    {{- else if .IsResponseByteStream -}}
+    InputStream response = {{ template "api-call" . }}
```
nfx, line 37: why not returning `InputStream` as a result type? it makes sense to return a wrapping response type only if there's more than one field in it, otherwise it's confusing.

.codegen/impl.java.tmpl
```diff
@@ -26,18 +28,36 @@ class {{.PascalName}}Impl implements {{.PascalName}}Service {
...
-                        {{- end -}}{{else}}, Void{{end}}.class);{{end}}
-    {{- else -}}apiClient.{{.Verb}}(path{{if .Request}}, request{{end}}{{if .Response -}}, {{if .Response.ArrayValue }}Collection
-      {{- else if .Response.MapValue }}Map
-      {{- else}}{{template "type" .Response}}
-      {{- end -}}{{else}}, Void{{end}}.class);
-    {{end}}
+    {{ template "headers" . -}}
+    {{ if not .Response -}}
+    {{ template "api-call" . }}
+    {{- else if .Response.ArrayValue -}} return apiClient.getCollection(path, null, {{template "type" .Response.ArrayValue}}.class, headers);
```
nfx, line 34: why this cannot be solved with overloading?
```suggestion
    {{- else if .Response.ArrayValue -}} return apiClient.getCollection(path, null, {{template "type" .Response.ArrayValue}}.class{{.HasRequestReaders}}, headers{{end}});
```

nfx (commented):
the diff seems to be bigger than necessary just for the addition of request headers.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
```diff
@@ -84,8 +84,7 @@ private Response computeResponse(Request in, CloseableHttpResponse response) thr
     boolean streamResponse =
         in.getHeaders().containsKey("Accept")
             && !APPLICATION_JSON.getMimeType().equals(in.getHeaders().get("Accept"))
-            && hs.containsKey("Content-Type")
```
mgyucht, line 87: This is needed because header keys are not case-sensitive (esp. if response is HTTP 2.0, which enforces header keys to be lower case).

tanmay-db (approved):
Thanks for going over the PR for review. LGTM

PR #140, opened 2023-08-19
Fix query param serialization for requests with enums

mgyucht:
## Changes
Query parameters are determined reflectively from request classes by scanning for fields with a QueryParam annotation. Serialization of query parameters is recursive in order to support complex types like the filter structures used for listing in the SQL query history service. This PR makes the following changes:

1. Terminate recursion when the request field is an enum.
2. When iterating through the request object's fields, skip any fields not annotated with QueryParam (but recursively, all fields do need to be serialized).
3. Rename the inner class from HeaderEntry to QueryParamPair (it represents query param pairs, not header entries).

## Tests
Test for this change is dependent on other testing refactors that will be merged as part of https://github.com/databricks/databricks-sdk-java/pull/139.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/GrpcTranscodingQueryParamsSerializer.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/GrpcTranscodingQueryParamsSerializer.java
@@ -17,11 +17,11 @@ import java.util.*;
  * documentation for gRPC transcoding</a> for more details.
  */
 public class GrpcTranscodingQueryParamsSerializer {
-  public static class HeaderEntry {
+  public static class QueryParamPair {
     private final String key;
     private final String value;
 
-    public HeaderEntry(String key, String value) {
+    public QueryParamPair(String key, String value) {
       this.key = key;
       this.value = value;
     }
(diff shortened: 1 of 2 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/GrpcTranscodingQueryParamsSerializer.java
```diff
@@ -103,24 +97,30 @@ private static String getFieldName(Field f) {
     }
   }
 
-  private static Map<String, Object> flattenObject(Object o) {
+  private static Map<String, Object> flattenObject(Object o, Boolean onlyAnnotatedFields) {
```
tanmay-db, line 100: Since the test is dependent on another PR, should we add a unit tests for this?

tanmay-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GrpcTranscodingQueryParamsSerializer.java
mgyucht: I could have... but the other PR should be merged in the next day or two, so I think we can deal with that.

PR #144, opened 2023-08-30
[PECO-1020] Adding mock constructor method for WorkspaceClient to support mocked ApiClient

gopalldb:
Adding additional constructor in WorkspaceClient to support mocking ApiClient

## Changes
A new constructor in WorkspaceClient to support mocking ApiClient

## Tests
Ran $mvn clean package to verify all test cases and compilation errors


mgyucht:
Thanks for contributing this! Note that WorkspaceClient.java is a generated file. You need to make this change in .codegen/workspace.java.tmpl as well, otherwise this change will be erased the next time we update the SDK.

gopalldb:
> Thanks for contributing this! Note that WorkspaceClient.java is a generated file. You need to make this change in .codegen/workspace.java.tmpl as well, otherwise this change will be erased the next time we update the SDK.

Done, Thanks for pointing out.

gopalldb:
jenkins merge

PR #149, opened 2023-09-04
Update version in pom.xml to 0.8.0

pietern:


pietern:
@tanmay-db Are there other places where the version needs to be bumped?

pietern:
@tanmay-db Thanks, your suggestions sound good. Can you make sure to track it some placE?

PR #153, opened 2023-09-14
Add ErrorInfo to API errors

hectorcast-db:
## Changes
Databricks API can return error details in case of errors. In some cases, users need to access such details to be able to solve the issue. This is the case for the errors of type ErrorInfo used by the Settings Platform.
This PR adds the necessary fields to the APIError to Unmarshal the ErrorDetails and to expose the details of type ErrorInfo to the users.

## Tests
- [X] Unit test
- [X] `make fmt`
- [X] Integration test

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksError.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksError.java
@@ -117,12 +117,8 @@ public class DatabricksError extends DatabricksException {
     return false;
   }
 
-  private List<ErrorDetail> getErrorsOfType(String type) {
+  List<ErrorDetail> getErrorsOfType(String type) {
     return this.details.stream().filter(e -> e.getType().equals(type)).collect(Collectors.toList());
-  }
-
-  public List<ErrorDetail> getErrorInfo() {
-    return this.getErrorsOfType(ERROR_INFO_TYPE);
   }
 
   private static boolean isCausedBy(Throwable throwable, Class<? extends Throwable> clazz) {
@@ -133,12 +129,5 @@ public class DatabricksError extends DatabricksException {
       return true;
     }
     return isCausedBy(throwable.getCause(), clazz);
-  }
-
-  public static List<ErrorDetail> getErrorInfo(Throwable throwable) {
-    if (!(throwable instanceof DatabricksError)) {
-      return Collections.emptyList();
-    }
-    return ((DatabricksError) throwable).getErrorInfo();
   }
 }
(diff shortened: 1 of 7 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksError.java
```diff
@@ -17,6 +20,8 @@
  * unrecoverable way and this exception should be thrown, potentially wrapped in another exception.
  */
 public class DatabricksError extends DatabricksException {
+
```
pietern, line 23: Newline

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksError.java
```diff
@@ -17,6 +20,8 @@
...
+  static final String ERROR_INFO_TYPE = "type.googleapis.com/google.rpc.ErrorInfo";
```
pietern, line 24: This can be private right?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksException.java
```diff
@@ -8,4 +14,11 @@ public DatabricksException(String message) {
   public DatabricksException(String message, Throwable cause) {
     super(message, cause);
   }
+
+  public List<ErrorDetail> getErrorInfo() {
+    if (!(this instanceof DatabricksError)) {
+      return Collections.emptyList();
+    }
+    return ((DatabricksError) this).getDetailsByType(ERROR_INFO_TYPE);
+  }
```
pietern, line 23: This accessor can live in `DatabricksError`. The caller should catch `DatabricksError` and work with it directly. This error being thrown means the Databricks backend actually responded.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/ErrorDetail.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  public void setMetadata(Map<String, String> metadata) {
+    this.metadata = metadata;
+  }
+
+  public String getDomain() {
+    return domain;
+  }
+
```
pietern, line 63: Do we need all of these setters?

hectorcast-db: Probably not. We have some setters in some of the error classes so I add them here. But better to only add them if needed. Removing.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksException.java
hectorcast-db: Moved

pietern (approved):

PR #157, opened 2023-09-26
Chunk streaming request bodies only

### LINKED ISSUE #154: [ISSUE] oauth-m2m fails to retrieve access token
**Description**
Hi, from version `0.7.0` up to the latest version (`0.8.1`), the SDK fails to retrieve an access token for oauth-m2m ClientCredentials. Version `0.6.0` authenticates just fine.

**Reproduction**

I just created a new `WorkspaceClient` with an explicit `DatabricksConfig`, and made a call like so:
```
DatabricksConfig config = new DatabricksConfig()
    .setHost("https://dbc-xxx-xxx.cloud.databricks.com")
    .setClientId("xxx-xxx")
    .setClientSecret("xxxxxx");
WorkspaceClient client = new WorkspaceClient(config);
client.catalogs().list();
```

This results in the following exception:
```
com.databricks.sdk.core.DatabricksException: oauth-m2m auth: Failed to refresh credentials: No content to map due to end-of-input
```

This appears to be caused by the token endpoint returning a `400 Bad Request` with an empty body. Having looked at the request that is passed to the Apache Http Client, I unfortunately didn't see anything wrong with it at first glance. Using `0.6.0` solves this issue however.

**Expected behavior**
The ClientCredentials should retrieve a token so that the workspace client can successfully perform requests.

**Debug Logs**
N.a.

**Other Information**
 - OS: macOS
 - Version: 0.7.0-0.8.1

**Additional context**
N.a.

Thanks, I hope you can reproduce it! 😊

mgyucht:
## Changes
#138 changed the CommonsHttpClient implementation to use `InputStreamEntity` for all request bodies sent using the Java SDK. As a result, requests are sent with a `Transfer-Encoding: chunked` header, and the request body is chunked accordingly. However, the Databricks REST API only tolerates chunked requests for certain APIs; for others, it ignores the request body if Content-Length is not specified. Auth falls under that category, which caused #154.

To fix this, Request will keep track of string entities and input stream entities separately. Previously, Request had separate fields for the body and debugBody, but they were both always set, so it wasn't possible to distinguish between these two cases. Now, HTTP clients can have different behavior based on whether the request body can be fully materialized as a string or is lazily read as with input streams.

As part of this, I have removed `SimpleHttpServer`. This was used before we were sure whether it was possible to use the HttpServer built into the JDK. Now that we are confident that we can (see usage in ExternalBrowserCredentialsProvider.java), I've replaced it with HttpServer and a custom HttpHandler in FixtureServer (the only place where it is used now).

One other small change: I've updated the logging configuration and dependency for the cli auth demo app. This ensures that users can see debug output (which I used when debugging this).

Closes #154. 

## Tests
I've refactored FixtureServer to support more advanced validation on the HTTP requests sent by clients for testing purposes. Now, users can assert that a request has a specific method, path, headers, body, and check for headers that should not be present.

One downside of this change is that users need to call the `.with()` method once per API call, rather than one time, at the start of each test.

Unit tests cover both cases (request body should be chunked when specified as a String; request body should not be chunked when specified as an InputStream).


databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
```diff
@@ -117,18 +122,26 @@ private HttpUriRequest transformRequest(Request in) {
...
-  private HttpRequestBase withEntity(HttpEntityEnclosingRequestBase request, InputStream body) {
-    request.setEntity(new InputStreamEntity(body));
+  private HttpRequestBase withEntity(HttpEntityEnclosingRequestBase request, Request in) {
+    if (in.isBodyString()) {
+      request.setEntity(new StringEntity(in.getBodyString(), StandardCharsets.UTF_8));
+    } else if (in.isBodyStreaming()) {
+      request.setEntity(new InputStreamEntity(in.getBodyStream()));
+    } else {
+      LOG.warn(
+          "withEntity called with a request with no body, so no request entity will be set. URI: {}",
```
tanmay-db, line 142: Right now, `withEntity` is only being called in POST / PUT and PATCH right? Is this pathway only for the PUT?

tanmay-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
mgyucht: `withEntity` should be called for any HTTP methods that have a request body, i.e. POST/PUT/PATCH. GET/DELETE and other methods we may support in the future like HEAD/OPTIONS don't support request bodies, so we don't use this method.

PR #160, opened 2023-10-02
Update OpenAPI spec

mgyucht:
## Changes
Update of the OpenAPI spec. This also introduces the template_libraries field being added in https://github.com/databricks/databricks-sdk-go/pull/635.

This PR contains some other changes:
1. Change `float`/`Float` to `double`/`Double`. OpenAPI specs currently specify `format: "double"` for all non-integral numeric values.
2. Remove old types that were previously in the OpenAPI spec but are no longer needed.
3. Do not generate files for types that don't correspond to distinct Java types (e.g. are not objects or enums). A lot of extra files are generated unnecessarily today, corresponding to either integers/strings/booleans or arrays/maps. Since Java has no type-def functionality, these are represented in the SDK as their JDK equivalents.

## Tests
Added AccountMetastoreAssignmentsIT to verify the correct functionality of the updated AccountMetastoreAssignments API.
Nightly tests passed on this PR.


.codegen/lib.tmpl
```diff
@@ -0,0 +1,26 @@
+{{- define "type-unboxed" -}}
+  {{- if .IsBool}}boolean
+  {{- else if .IsInt64}}long
+  {{- else if .IsFloat64}}double
```
tanmay-db, line 4: good catch

databricks-sdk-java/src/main/java/com/databricks/sdk/service/provisioning/CustomTags.java
```diff
@@ -1,10 +0,0 @@
-// Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.
-
-package com.databricks.sdk.service.provisioning;
-
-/**
- * The custom tags key-value pairing that is attached to this workspace. The key-value pair is a
- * string of utf-8 characters. The value can be an empty string, with maximum length of 255
- * characters. The key can be of maximum length of 127 characters, and cannot be empty.
- */
-// type CustomTags Map<String,String>
```
tanmay-db, line 10: Since these were never used, this change shouldn't be breaking right?

tanmay-db (approved):

.codegen/model.java.tmpl
```diff
@@ -75,23 +73,9 @@ public enum {{.PascalName}}{
   @JsonProperty("{{.Content}}") {{end}}{{.ConstantName}},{{.Comment "// " 80}}
   {{ end }}
 }
+{{ else }}
+Any types that have no fields (i.e. primitives or array types) or are not enums are not represented by distinct Java
+types in the Java SDK.
+{{skipThisFile}}
```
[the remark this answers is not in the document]
mgyucht: Correct. That's why several files disappear from the .gitattributes file.

databricks-sdk-java/src/main/java/com/databricks/sdk/service/provisioning/CustomTags.java
mgyucht: Correct.

PR #170, opened 2023-10-18
Adjust token expiry window to 40 seconds because of Azure

tanmay-db:
## Changes

- Align with https://github.com/databricks/databricks-sdk-py/pull/392
- Add ClockSupplier so that we don't depend on system clock while testing. Modify Token class to be more test friendly. 

## Tests

Unit Tests
All nightly tests passed except two tests (which failed due to Max limit of 100 scopes has been reached!). It's a known current test infra issue and unrelated to these changes.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java
@@ -30,7 +30,7 @@ public class Token {
     this(accessToken, tokenType, null, expiry, null);
   }
 
-  /** Constructor for non-refreshable tokens (e.g. M2M). with ClockSupplier */
+  /** Constructor for non-refreshable tokens (e.g. M2M) with ClockSupplier */
   public Token(
       String accessToken, String tokenType, LocalDateTime expiry, ClockSupplier clockSupplier) {
     this(accessToken, tokenType, null, expiry, clockSupplier);
@@ -41,7 +41,7 @@ public class Token {
     this(accessToken, tokenType, refreshToken, expiry, null);
   }
 
-  /** Constructor for refreshable tokens. with ClockSupplier */
+  /** Constructor for refreshable tokens with ClockSupplier. */
   public Token(
       String accessToken,
       String tokenType,
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java
@@ -41,7 +41,9 @@ public class Token {
     if (expiry == null) {
       return false;
     }
-    LocalDateTime potentiallyExpired = expiry.minus(10, ChronoUnit.SECONDS);
+    // Azure Databricks rejects tokens that expire in 30 seconds or less,
+    // so we refresh the token 40 seconds before it expires.
+    LocalDateTime potentiallyExpired = expiry.minus(40, ChronoUnit.SECONDS);
     LocalDateTime now = LocalDateTime.now();
     return potentiallyExpired.isBefore(now);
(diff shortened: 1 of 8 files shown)


pietern (approved):
Nice work! Let's wait for @mgyucht to have a look as well.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java
```diff
@@ -21,28 +23,51 @@ public class Token {
    */
   @JsonProperty private LocalDateTime expiry;
 
+  private final ClockSupplier clockSupplier;
+
   /** Constructor for non-refreshable tokens (e.g. M2M). */
   public Token(String accessToken, String tokenType, LocalDateTime expiry) {
-    this(accessToken, tokenType, null, expiry);
+    this(accessToken, tokenType, null, expiry, null);
```
mgyucht, line 30: Slightly simpler would be to pass `new RealClockSupplier()` here, then you don't need an if condition in the other case, and you can use `Objects.requireNonNull`.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/RealClockSupplier.java
```diff
@@ -0,0 +1,10 @@
+package com.databricks.sdk.core.utils;
+
+import java.time.Clock;
+
+public class RealClockSupplier implements ClockSupplier {
```
mgyucht, line 5: nit: RealClockSupplier is a bit of a funny name, like as opposed to imaginary? It also doesn't say what the source of the clock is. How about SystemClockSupplier?

mgyucht (approved):
Two small nits. Thanks, feel free to address and merge

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java
tanmay-db: @mgyucht thanks for review. That would be bad CUJ for someone calling the other constructor right? They would have to pass Clock Supplier which outside of test should always be System Clock. 

The clock shouldn't be exposed on customer end I think.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/RealClockSupplier.java
tanmay-db: Yes, I will update it. I took the inspiration for naming from RealTimer and FakeTimer.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java
mgyucht: Isn't the other constructor only used for tests? If you want to use the system clock, you should use this constructor.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/RealClockSupplier.java
mgyucht: Ah that's a good point... yeah I think system is better.

tanmay-db: Yeah, I will update the RealTimer -> SystemTimer as well

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java
tanmay-db: Right, I was thinking if someone wants to pass null in clockSupplier -- we should have a case for that but they should then use the other constructor. Updated.

PR #171, opened 2023-10-23
Implement notebook-native auth for the Java SDK

mgyucht:
## Changes
Notebook-native authentication has been supported in the Python SDK for the last several months, and customers using the SDKs in Databricks appreciate the easier CUJ for authentication. This PR ports the behavior to the Java SDK. The approach taken for this implementation differs somewhat from the Python SDK's implementation. In particular, in Python, DBR provides some functionality to get the command context (`from dbruntime.databricks_repl_context import get_context`, `from dbruntime.sdk_credential_provider import init_runtime_native_auth`) as well as from IPython. The Java SDK uses the DBUtils object directly. However, the SDK doesn't depend on the DBUtils API (which is only in Scala), so instead we load it reflectively.

## Tests
Ran in a notebook in DBR, worked.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
@@ -2,23 +2,24 @@ package com.databricks.sdk.core;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.util.HashMap;
 import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A CredentialsProvider that uses the API token from the command context to authenticate.
- * <p>
- * The token and hostname are read from the command context, which can be retrieved through the dbutils API. As the Java
- * SDK does not depend on DBUtils directly, reflection is used to retrieve the token. This token should be available
- * wherever the DBUtils API is accessible (i.e. in the Spark driver).
+ *
+ * <p>The token and hostname are read from the command context, which can be retrieved through the
+ * dbutils API. As the Java SDK does not depend on DBUtils directly, reflection is used to retrieve
+ * the token. This token should be available wherever the DBUtils API is accessible (i.e. in the
+ * Spark driver).
  */
 public class NotebookNativeCredentialsProvider implements CredentialsProvider {
-  private static final Logger LOG = LoggerFactory.getLogger(NotebookNativeCredentialsProvider.class);
+  private static final Logger LOG =
+      LoggerFactory.getLogger(NotebookNativeCredentialsProvider.class);
   private static final ObjectMapper MAPPER = new ObjectMapper();
 
   @Override
@@ -33,7 +34,8 @@ public class NotebookNativeCredentialsProvider implements CredentialsProvider {
     }
 
     // DBUtils is not available in the Java SDK, so we have to use reflection to get the token.
-    // First, we get the context by calling getContext on the notebook field of dbutils, then we get the apiKey and
+    // First, we get the context by calling getContext on the notebook field of dbutils, then we get
+    // the apiKey and
     // apiUrl fields from the context. If this is successful, we set the host on the config.
     try {
       Object dbutils = getDbUtils();
@@ -49,7 +51,10 @@ public class NotebookNativeCredentialsProvider implements CredentialsProvider {
         TokenAndUrl tokenAndUrl;
         try {
           tokenAndUrl = getTokenAndUrl(notebook);
-        } catch (IllegalAccessException | NoSuchMethodException | InvocationTargetException | JsonProcessingException e) {
+        } catch (IllegalAccessException
+            | NoSuchMethodException
+            | InvocationTargetException
+            | JsonProcessingException e) {
           throw new RuntimeException(e);
         }
         headers.put("Authorization", String.format("Bearer %s", tokenAndUrl.token));
@@ -109,9 +114,17 @@ public class NotebookNativeCredentialsProvider implements CredentialsProvider {
   }
 
   /** Fetch the current command context, and read the API token and URL from it. */
-  private static TokenAndUrl getTokenAndUrl(Object notebook) throws NoSuchMethodException, InvocationTargetException, IllegalAccessException, JsonProcessingException {
-    Object testCommandContext = notebook.getClass().getDeclaredMethod("getContext").invoke(notebook);
-    String json = (String) testCommandContext.getClass().getDeclaredMethod("safeToJson").invoke(testCommandContext);
+  private static TokenAndUrl getTokenAndUrl(Object notebook)
+      throws NoSuchMethodException, InvocationTargetException, IllegalAccessException,
+          JsonProcessingException {
+    Object testCommandContext =
+        notebook.getClass().getDeclaredMethod("getContext").invoke(notebook);
+    String json =
+        (String)
+            testCommandContext
+                .getClass()
+                .getDeclaredMethod("safeToJson")
+                .invoke(testCommandContext);
     CommandContext deserialized = MAPPER.readValue(json, CommandContext.class);
     String token = deserialized.attributes.get("api_token");
     String host = deserialized.attributes.get("api_url");
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
@@ -1,0 +1,109 @@
+package com.databricks.sdk.core;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class NotebookNativeCredentialsProvider implements CredentialsProvider {
+  private static final Logger LOG = LoggerFactory.getLogger(NotebookNativeCredentialsProvider.class);
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  @Override
+  public String authType() {
+    return "runtime";
+  }
+
+  @Override
+  public HeaderFactory configure(DatabricksConfig config) {
+    if (System.getenv("DATABRICKS_RUNTIME_VERSION") == null) {
+      return null;
+    }
+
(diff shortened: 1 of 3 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
```diff
@@ -0,0 +1,133 @@
...
+ * Spark driver).
+ */
+public class NotebookNativeCredentialsProvider implements CredentialsProvider {
+  private static final Logger LOG =
+      LoggerFactory.getLogger(NotebookNativeCredentialsProvider.class);
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  @Override
+  public String authType() {
+    return "runtime";
```
tanmay-db, line 27: Thought -- I think `runtime` auth here might be confusing, do we want to have `notebook-native` auth type and alias that to `runtime`? But if we decide to go in this direction then this would need to be backported to python SDK as well and there probably isn't high ROI to do this.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * <p>The token and hostname are read from the command context, which can be retrieved through the
+ * dbutils API. As the Java SDK does not depend on DBUtils directly, reflection is used to retrieve
+ * the token. This token should be available wherever the DBUtils API is accessible (i.e. in the
+ * Spark driver).
+ */
+public class NotebookNativeCredentialsProvider implements CredentialsProvider {
+  private static final Logger LOG =
+      LoggerFactory.getLogger(NotebookNativeCredentialsProvider.class);
+  private static final ObjectMapper MAPPER = new ObjectMapper();
```
tanmay-db, line 23: I think there was a discussion sometime back on having a single object mapper but not sure if there was any conclusion regarding that.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
```diff
@@ -0,0 +1,133 @@
...
+
+  @Override
+  public HeaderFactory configure(DatabricksConfig config) {
+    if (System.getenv("DATABRICKS_RUNTIME_VERSION") == null) {
+      return null;
+    }
+
+    // DBUtils is not available in the Java SDK, so we have to use reflection to get the token.
+    // First, we get the context by calling getContext on the notebook field of dbutils, then we get
+    // the apiKey and
```
tanmay-db, line 38: nit: new line, bottom comment can continue after `apiKey and ...` rather than be in newline

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /** Load the dbutils object initialized by DBR. */
+  private static Object getDbUtils() {
+    try {
+      Class<?> dbutilsHolderClass = Class.forName("com.databricks.dbutils_v1.DBUtilsHolder$");
+      Object dbutilsHolder = dbutilsHolderClass.getDeclaredField("MODULE$").get(null);
+      InheritableThreadLocal<Object> dbutils = getField(dbutilsHolder, "dbutils0");
+      return dbutils.get();
+    } catch (Exception e) {
```
tanmay-db, line 76: Catching a specific exception might be better

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
```diff
@@ -0,0 +1,133 @@
...
+        } catch (IllegalAccessException
+            | NoSuchMethodException
+            | InvocationTargetException
+            | JsonProcessingException e) {
+          throw new RuntimeException(e);
+        }
+        headers.put("Authorization", String.format("Bearer %s", tokenAndUrl.token));
+        return headers;
+      };
+    } catch (Exception e) {
```
tanmay-db, line 63: Same here about catching specific exception.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
```diff
@@ -0,0 +1,133 @@
...
+    if (System.getenv("DATABRICKS_RUNTIME_VERSION") == null) {
+      return null;
+    }
+
+    // DBUtils is not available in the Java SDK, so we have to use reflection to get the token.
+    // First, we get the context by calling getContext on the notebook field of dbutils, then we get
+    // the apiKey and
+    // apiUrl fields from the context. If this is successful, we set the host on the config.
+    try {
+      Object dbutils = getDbUtils();
```
tanmay-db, line 41: should we check for dbutils to be null before passing that to getFields?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
```diff
@@ -0,0 +1,133 @@
...
+      InheritableThreadLocal<Object> dbutils = getField(dbutilsHolder, "dbutils0");
+      return dbutils.get();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  /** Reflectively get a field by name from an object. */
+  private static <T> T getField(Object o, String fieldName) {
+    Field f;
```
tanmay-db, line 83: nit: fieldName / Object can be null but given we are only using this in scope of notebook native auth, I think it's okay if we don't have a check.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
```diff
@@ -0,0 +1,133 @@
...
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  @Override
+  public String authType() {
+    return "runtime";
+  }
+
+  @Override
+  public HeaderFactory configure(DatabricksConfig config) {
+    if (System.getenv("DATABRICKS_RUNTIME_VERSION") == null) {
```
tanmay-db, line 32: I think instead of checking this here, we should have an attribute in config which would be set at resolve() and then check that attribute here. We do that for other providers.

tanmay-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
mgyucht: This is already implemented in the Python SDK, and the auth type there is called `runtime`: https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/core.py#L95

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
mgyucht: Interesting. I was mainly copying the implementation from https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/core.py#L97-L98, which just reads this environment variable from the actual environment. I suppose we could add this to the configuration itself in all SDKs, need to think about it a bit more.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
mgyucht: True. I can throw an exception here if needed.

mgyucht:
Need to check DBR version range supported (including serverless)

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
mgyucht: There are ten currently in the SDK. That is something we can clean up, but would rather skip for now.

tanmay-db: Got it.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
mgyucht: Can you give a concrete example? The issue is that we actually do want to catch all exceptions, and we don't want to handle any exceptions differently. The codepath should always be "rethrow as DatabricksException", and then `configure()` can have a top-level path of just "log at debug and return null".

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
mgyucht: Thinking about this a bit more, I think we should keep this as-is. The main issue is that this environment variable is not expected to be configured or set by users, it's just a way that we detect whether we're running in DBR.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/NotebookNativeCredentialsProvider.java
tanmay-db: I need to check for the underlying methods for concrete exceptions we should catch but I think catching all exception and throwing them as DatabricksException isn't useful for example -- if there is NullPointerException (which is a RuntimeException) but we throw that wrapped in DatabricksException. It would pollute the stack trace and in most cases the NullPointerException happens because of a bug -- catching this would hide (to some extent) the underlying bug. 

> The codepath should always be "rethrow as DatabricksException", and then configure() can have a top-level path of just "log at debug and return null".
But we should do this only for specific exceptions right and not for all exceptions?

mgyucht: In the auth codepath, the contract is that `configure()` returns null to indicate that the credential provider can't be used with the given `DatabricksConfig`. Throwing an exception aborts the `configure()` method. Because of this, we explicitly do not want to throw exceptions across this error boundary. Instead, we can log exceptions as they are thrown at DEBUG so that users can see them if needed for debugging. Note that I'm never discarding the original exception; it is listed as the cause of the DatabricksException, so it will be visible in the `Caused by` section of the stack trace when printed out.

mgyucht: What specific exception would we want to handle differently, and how should we handle it?

mgyucht: That's fair. We would still catch exceptions broadly at the root. I can make sure to catch the checked exceptions and do some null checking. I just need to test this carefully to make sure this doesn't cause issues with stability (though, I guess it is the last provider in DefaultCredentialsProvider, so even if it fails, it won't block other providers from loading).

tanmay-db (approved):
LGTM

PR #174, opened 2023-10-25
Use same logic as Go SDK for DatabricksConfig.isAzure()

mgyucht:
## Changes
The current isAzure() logic only checks if ".azuredatabricks." is contained in the hostname, which fails for workspaces in other Azure environments. This PR copies the logic directly from the Go SDK: https://github.com/databricks/databricks-sdk-go/blob/main/config/config.go#L142

## Tests
- [x] Some unit tests ensuring that isAzure() returns the right values for various hosts, including AWS, Azure public, Azure US Gov, and Azure China.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
@@ -463,7 +463,7 @@ public class DatabricksConfig {
     if (host == null) {
       return false;
     }
-    return host.contains(".azuredatabricks.");
+    return host.contains(".azuredatabricks.") || host.contains("databricks.azure.cn") || host.contains(".databricks.azure.us");
   }
 
   public synchronized void authenticate(HttpMessage request) {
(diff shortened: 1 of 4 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -463,7 +463,9 @@ public boolean isAzure() {
     if (host == null) {
       return false;
     }
-    return host.contains(".azuredatabricks.");
+    return host.contains(".azuredatabricks.net")
```
tanmay-db, line 466: Curious -- why do we need to limit it to `.net`?

tanmay-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
mgyucht: I'm just matching the Go SDK logic here.

tanmay-db: I see

PR #177, opened 2023-11-01
Mark Java SDK as Beta

mgyucht:
## Changes
After discussion, we decided that we're ready to mark the Java SDK at beta stability, much like the Python and Go SDKs.


tanmay-db (approved):
🎉

PR #185, opened 2023-11-10
Introduce more specific exceptions, like `NotFound`, `AlreadyExists`, `BadRequest`, `PermissionDenied`, `InternalError`, and others

nfx:
See implementations in other SDKs:

- Go: https://github.com/databricks/databricks-sdk-go/pull/682
- Python: https://github.com/databricks/databricks-sdk-py/pull/376

### DIFF
--- a/.codegen/exception.java.tmpl
+++ b/.codegen/exception.java.tmpl
@@ -1,0 +1,24 @@
+// Code generated from OpenAPI specs by Databricks SDK Generator. DO NOT EDIT.
+
+package com.databricks.sdk.core.error;
+
+import com.databricks.sdk.core.DatabricksError;
+import com.databricks.sdk.support.Generated;
+
+import java.util.List;
+
+/**
+ {{.Comment " * " 80}}
+ */
+@Generated
+public class {{.PascalName}} extends {{if .Inherit -}}{{.Inherit.PascalName}}{{else}}DatabricksError{{end}} {
+  public {{.PascalName}}(String message, List<ErrorDetail> details) {
+    super("{{.Name}}", message, {{.StatusCode}}, details);
+  }
+
+  {{if not .Inherit}}
+  {{.PascalName}}(String errorCode, String message, int statusCode, List<ErrorDetail> details) {
+      super(errorCode, message, statusCode, details);
+  }
+  {{end}}
+}
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/AbstractErrorMapper.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/AbstractErrorMapper.java
@@ -1,0 +1,37 @@
+package com.databricks.sdk.core.error;
+
+import com.databricks.sdk.core.DatabricksError;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+abstract class AbstractErrorMapper {
+  @FunctionalInterface
+  protected interface Rule {
+    DatabricksError create(String message, List<ErrorDetail> details);
+  }
+
+  public DatabricksError apply(int code, ApiErrorBody errorBody) {
+    String message = errorBody.getMessage();
+    String errorCode = errorBody.getErrorCode();
+    List<ErrorDetail> details = errorBody.getErrorDetails();
+    if (statusCodeMapping.containsKey(code)) {
+      return statusCodeMapping.get(code).create(message, details);
+    }
+    if (errorCodeMapping.containsKey(errorCode)) {
+      return errorCodeMapping.get(errorCode).create(message, details);
+    }
+    return new DatabricksError(errorCode, message, code, details);
+  }
+
+  private final Map<Integer, Rule> statusCodeMapping = new HashMap<>();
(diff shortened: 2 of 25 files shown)


.codegen/exception.java.tmpl
```diff
@@ -0,0 +1,24 @@
...
+public class {{.PascalName}} extends {{if .Inherit -}}{{.Inherit.PascalName}}{{else}}DatabricksError{{end}} {
+  public {{.PascalName}}(String message, List<ErrorDetail> details) {
+    super("{{.Name}}", message, {{.StatusCode}}, details);
+  }
+
+  {{if not .Inherit}}
+  {{.PascalName}}(String errorCode, String message, int statusCode, List<ErrorDetail> details) {
+      super(errorCode, message, statusCode, details);
+  }
+  {{end}}
```
mgyucht, line 23: This case isn't used yet, but I suspect you may want to use it if you follow my comments in AbstractErrorMapper.java.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/AbstractErrorMapper.java
```diff
@@ -0,0 +1,37 @@
...
+  public DatabricksError apply(int code, ApiErrorBody errorBody) {
+    String message = errorBody.getMessage();
+    String errorCode = errorBody.getErrorCode();
+    List<ErrorDetail> details = errorBody.getErrorDetails();
+    if (statusCodeMapping.containsKey(code)) {
+      return statusCodeMapping.get(code).create(message, details);
+    }
+    if (errorCodeMapping.containsKey(errorCode)) {
+      return errorCodeMapping.get(errorCode).create(message, details);
+    }
```
mgyucht, line 23: I think this is in the wrong order: we should check for specific error code first and then fall back to status code.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/AbstractErrorMapper.java
```diff
@@ -0,0 +1,37 @@
...
+  protected interface Rule {
+    DatabricksError create(String message, List<ErrorDetail> details);
+  }
+
+  public DatabricksError apply(int code, ApiErrorBody errorBody) {
+    String message = errorBody.getMessage();
+    String errorCode = errorBody.getErrorCode();
+    List<ErrorDetail> details = errorBody.getErrorDetails();
+    if (statusCodeMapping.containsKey(code)) {
+      return statusCodeMapping.get(code).create(message, details);
```
mgyucht, line 19: I think this could cause some confusing behavior for old clients when new error codes are introduced. The error code provided in the API response is not explicitly propagated to these exception types. I think you might need two "rule" interfaces: one for ErrorCode-based exceptions (with well-known error codes, so only need parameters for message and details) and one for StatusCode-based exceptions (with dynamic error-codes, so you need an extra parameter for the error code).

mgyucht (commented):
Largely looks good to me, but I think there are a couple issues with ordering in ErrorMapper and with error codes being dropped for status code-based errors.

databricks-sdk-java/src/test/java/com/databricks/sdk/integration/ClustersIT.java
```diff
@@ -47,4 +48,14 @@ void latestRuntime(WorkspaceClient w) {
 
     assertNotNull(runtime);
   }
+
+  @Test
+  void clusterDoesNotExist(WorkspaceClient w) {
+    assertThrowsExactly(
```
mgyucht, line 54: This test will fail once I merge the follow-up change (equivalent of  https://github.com/databricks/databricks-sdk-py/commit/45a356ca24ce92c80f06c3f60f9969aaf25987e4)

edwardfeng-db: ack, thanks

databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/AbstractErrorMapper.java
```diff
@@ -0,0 +1,42 @@
...
+  public DatabricksError apply(int code, ApiErrorBody errorBody) {
+    String message = errorBody.getMessage();
+    String errorCode = errorBody.getErrorCode();
+    List<ErrorDetail> details = errorBody.getErrorDetails();
+    if (errorCodeMapping.containsKey(errorCode)) {
+      return errorCodeMapping.get(errorCode).create(message, details);
+    }
+    if (statusCodeMapping.containsKey(code)) {
+      return statusCodeMapping.get(code).create(errorCode, message, details);
+    }
```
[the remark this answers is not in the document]
mgyucht: Error codes always take precedence over status codes. The same status code could be used to represent different error codes. Because of this, we don't need to store status code there.

databricks-sdk-java/src/test/java/com/databricks/sdk/integration/ClustersIT.java
mgyucht: I will fix it in that PR, no worries.l

PR #190, opened 2023-11-17
Support custom scopes and redirectUrl for U2M OAuth flow

gopalldb:
## Changes

Added following utility methods to DatabricksConfig

a. setRedirectUrl: This will be passed to OAuthClient and will be used in U2M OAuth flow if client wants to provide a custom redirect Url as per OAuth configuration of their application.
b. setScopes: Clients can set custom scopes for U2M OAuth flow. If offline_access is not present, it will be added to the scopes for refresh token support.

## Tests

Tested manually using JDBC driver as a client. Also, added test case for custom values of redirectUrl and scopes.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
@@ -7,6 +7,7 @@ import com.databricks.sdk.core.http.Response;
 import com.databricks.sdk.core.oauth.OpenIDConnectEndpoints;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.Supplier;
@@ -36,6 +37,20 @@ public class DatabricksConfig {
       auth = "oauth",
       sensitive = true)
   private String clientSecret;
+
+  @ConfigAttribute(
+      value = "scopes",
+      env = "DATABRICKS_SCOPES",
+      auth = "oauth",
+      sensitive = true)
+  private List<String> scopes;
+
+  @ConfigAttribute(
+      value = "redirect_url",
+      env = "DATABRICKS_REDIRECT_URL",
+      auth = "oauth",
+      sensitive = true)
+  private String redirectUrl;
 
   @ConfigAttribute(value = "username", env = "DATABRICKS_USERNAME", auth = "basic")
   private String username;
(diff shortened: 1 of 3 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -37,6 +38,16 @@ public class DatabricksConfig {
       sensitive = true)
   private String clientSecret;
 
+  @ConfigAttribute(value = "scopes", env = "DATABRICKS_SCOPES", auth = "oauth", sensitive = true)
```
mgyucht, line 41: Scopes are not sensitive, are they?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -37,6 +38,16 @@ public class DatabricksConfig {
...
+  private List<String> scopes;
+
+  @ConfigAttribute(
+      value = "redirect_url",
+      env = "DATABRICKS_REDIRECT_URL",
+      auth = "oauth",
+      sensitive = true)
```
mgyucht, line 48: Ditto, probably not sensitive.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/OAuthClient.java
```diff
@@ -116,6 +120,9 @@ private OAuthClient(Builder b) throws IOException {
     List<String> scopes = b.scopes;
     if (scopes == null) {
       scopes = Arrays.asList("offline_access", "clusters", "sql");
+    } else if (!scopes.contains("offline_access")) {
+      scopes = new ArrayList<>(scopes);
+      scopes.add("offline_access");
```
mgyucht, line 125: Why should we add this always?

mgyucht (commented):
Overall seems good, but I am not sure why we would want all apps to request offline_access scope, it seems like it should be opt-in like with the /token API itself.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
gopalldb: The validation logic is case sensitive: https://src.dev.databricks.com/databricks/universe/-/blob/common/authentication/oauth/OAuthAuthorizationEndpointHandler.scala?L91:41

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/OAuthClient.java
gopalldb: Removed. Added to support refresh tokens, but clients may intentionally avoid this for short lived token use-case.

gopalldb:
> Overall seems good, but I am not sure why we would want all apps to request offline_access scope, it seems like it should be opt-in like with the /token API itself.

Removed that. Now it is opt-in for clients.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
mgyucht: Sensitive here means that it shouldn't be put into logs, not that it is case-sensitive.

gopalldb: My bad. I misunderstood. :( Updated

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
gopalldb: Updated

mgyucht (approved):

PR #192, opened 2023-11-23
Fix databricks CLI authentication on Windows

mgyucht:
## Changes
There are two problems with respect to the Java SDK's auth pathways:
1. The logic to find the `.databrickscfg` file only works on *nix, as it hard-codes the path using forward-slashes. I've rewritten this to use `Paths.get`.
2. The databricks binary on Windows is called `databricks.exe`, not `databricks`, so we need to search for the binary with OS-specific filename.

To make this change unit-testable, I introduce an `Environment` class to abstract over environment variables, the search path for the databricks CLI executable, and the operating system name. This is needed because there are both differences in the environment variable name (`PATH` in Linux/Mac vs `Path` in Windows) and in the path separator (`:` in Linux/Mac vs `;` in Windows). The only system-independent way to get the path seems to be `System.getenv("PATH")` which works even on Windows. Note that `System.getenv().get("PATH")` doesn't work on Windows. Additionally, determining the OS depends on `System.getProperty`, so this can be initialized directly in DatabricksConfig and mocked in unit tests.

## Tests
Manually tested on a Windows machine:
1. Installed the Databricks CLI and ran `.\datarbricks.exe auth login --host <HOST>`.
2. Added the Databricks CLI to the PATH system environment variable.
3. Ran the CLIWorkspaceAuth example and was able to list clusters.

As a side note: when we make auth permutations test harnesses, we should have a Windows variant to catch these errors.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/OSUtils.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/OSUtils.java
@@ -23,10 +23,7 @@ public class OSUtils {
    *     for Windows, "mac" for macOS, and "linux" for Linux-based operating systems.
    */
   static String getOS(Environment env) {
-    String systemName = env.get("OS_NAME");
-    if (systemName == null || systemName.isEmpty()) {
-      systemName = System.getProperty("os.name").toLowerCase();
-    }
+    String systemName = env.getSystemName();
     if (systemName.startsWith("win")) {
       return "win";
     } else if (systemName.startsWith("mac")) {
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/OSUtils.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/OSUtils.java
@@ -42,17 +42,17 @@ public class OSUtils {
     String os = getOS(cfg);
     switch (os) {
       case "win":
-        return new WindowsUtilities();
+        return new WindowsUtilities(cfg);
       case "mac":
-        return new MacOSUtilities();
+        return new MacOSUtilities(cfg);
       case "linux":
-        return new LinuxUtilities();
+        return new LinuxUtilities(cfg);
     }
     throw new DatabricksException("Unsupported OS: " + os);
   }
 
-  public static String findExecutable(String name) {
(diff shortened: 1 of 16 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/OSUtils.java
```diff
@@ -1,44 +1,74 @@
...
+      }
+      if (size < 1024 * 1024) {
+        LOG.debug("Databricks CLI version <0.100.0 detected");
+        continue;
+      }
+      String pathStr = path.toString();
+      LOG.debug("Found executable named '" + name + "' in PATH: " + pathStr);
+      return pathStr;
     }
+    LOG.debug("Failed to find executable named '" + name + "' in PATH");
```
tanmay-db, line 71: Since we are mentioning failed in the message -- this probably should be LOG.warn

examples/cli-auth-demo/src/main/java/com/databricks/example/CliAuthWorkspace.java
```diff
@@ -30,9 +30,9 @@ private static DatabricksConfig getConfig() {
      Authenticate and retrieve the list of clusters from the workspace
      */
     public static void main(String[] args) {
-        DatabricksConfig config = getConfig();
+        // DatabricksConfig config = getConfig();
```
tanmay-db, line 33: Should we remove this comment from demo?

tanmay-db (approved):
Thanks for the fix. LGTM

examples/cli-auth-demo/src/main/java/com/databricks/example/CliAuthWorkspace.java
mgyucht: Right, thanks for the reminder, I was going to undo this change

databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/OSUtils.java
mgyucht: I don't think this is an indication that there is a problem though. The user may legitimately not have installed the CLI in their system (e.g. on a webserver).

PR #193, opened 2023-11-24
Fix SCIM pagination

mgyucht:
## Changes
Supersedes https://github.com/databricks/databricks-sdk-java/pull/188.

This PR incorporates two hard-coded changes for the SCIM API in the Python SDK:

startIndex starts at 1 for SCIM APIs, not 0. However, the existing .Pagination.Increment controls both the start index as well as whether the pagination is per-page or per-resource. Later, we should replace this extension with two independent OpenAPI options: one_indexed (defaulting to false) and pagination_basis (defaulting to resource but can be overridden to page).
If users don't specify a limit, the SDK will include a hard-coded limit of 100 resources per request. We could add this to the OpenAPI spec as an option default_limit, which is useful for any non-paginated APIs that later expose pagination options and allow the SDK to gracefully support those. However, we don't want to encourage folks to use this pattern: all new list APIs are required to be paginated from the start.


.codegen/api.java.tmpl
```diff
@@ -131,7 +131,18 @@ public class {{.PascalName}}API {
...
+    request.set{{.Pagination.Offset.PascalName}}(
+    {{- if eq .Pagination.Increment 1 -}}
+      1
+    {{- else if contains .Path "/scim/v2/" -}}
+      1
+    {{- else -}}
+      0
+    {{- end}}L);{{end -}}
+    {{if and .Pagination.Limit (contains .Path "/scim/v2/")}}
+    if (request.get{{.Pagination.Limit.PascalName}}() == 0) {
```
tanmay-db, line 143: nit: 0L instead of 0. It doesn't matter because implicitly java will know 0 is of Long type. But mentioning since we specify 1L and 100L below so might be good for consistency on code reading purposes.

tanmay-db (approved):

PR #196, opened 2023-11-30
Support GCP auth

edwardfeng-db:
## Changes

- Make java-sdk support gcp auth

## Tests

- Integration tests
- Unit tests
- Manual tests

### DIFF
--- a/databricks-sdk-java/src/test/java/com/databricks/sdk/integration/CredentialsIT.java
+++ b/databricks-sdk-java/src/test/java/com/databricks/sdk/integration/CredentialsIT.java
@@ -15,7 +15,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
 public class CredentialsIT {
   @Test
   void lists(AccountClient a) {
-    // Skipping this test for gcp because there is a permission issue with listing credentials.
+    // Skipping this test for GCP because this api is not enabled in GCP.
     if (!a.config().isGcp()) {
       Iterable<Credential> list = a.credentials().list();
 


databricks-sdk-java/src/test/java/com/databricks/sdk/integration/FilesIT.java
```diff
@@ -46,10 +45,6 @@ private void writeFileAndReadFileInner(
     // Write the file to DBFS.
     workspace.files().upload(fileName, inputStream);
 
-    // Read file status
-    FileInfo fileStatus = workspace.files().getStatus(fileName);
-    Assertions.assertEquals(fileStatus.getPath(), fileName);
-
```
edwardfeng-db, line 52: Get status doesn't work. Same reason as for the python sdk [here](https://github.com/databricks/databricks-sdk-py/pull/444/files#diff-a4a7d7aa2604f48fbe447a1274bcf9aeb9d1805e4087ac4d036d8fd6d44ff230L225-L227), removing these two lines.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+      }
+    }
+
+    List<Option> tokenOption = Collections.emptyList();
+
+    ServiceAccountCredentials finalServiceAccountCredentials = serviceAccountCredentials;
+    return () -> {
+      IdToken idToken;
+      try {
+        idToken = finalServiceAccountCredentials.idTokenWithAudience(host, tokenOption);
```
mgyucht, line 57: We need to check whether we can request the ID token and SP access token before returning. If these fail during `configure()` time, we can return `null` and log a warning, allowing the user to fall back to another credential provider. If that succeeds, then we do this here.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+      } catch (IOException e) {
```
mgyucht, line 58: What specific exceptions are thrown here?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+        LOG.warn("Failed to get id token from Google service account credentials." + e);
+        throw new RuntimeException(e);
```
mgyucht, line 60: Let's throw DatabricksException instead.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+      }
+      Map<String, String> headers = new HashMap<>();
+      headers.put("Authorization", String.format("Bearer %s", idToken.getTokenValue()));
+
+      if (config.isAccountClient()) {
+        AccessToken token;
+        try {
+          token = finalServiceAccountCredentials.createScoped(GcpScopes).refreshAccessToken();
```
mgyucht, line 68: Ditto, we should check whether we can even fetch this token before returning so we can return `null` if it fails.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+        } catch (Exception e) {
```
mgyucht, line 69: Ditto, what exceptions are thrown here?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+          LOG.warn("Failed to refresh access token from Google service account credentials." + e);
+          throw new RuntimeException(e);
```
mgyucht, line 71: Ditto, let's use DatabricksException

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+
+    List<Option> tokenOption = Collections.emptyList();
+
+    ServiceAccountCredentials finalServiceAccountCredentials = serviceAccountCredentials;
+    return () -> {
+      IdToken idToken;
+      try {
+        idToken = finalServiceAccountCredentials.idTokenWithAudience(host, tokenOption);
+      } catch (IOException e) {
+        LOG.warn("Failed to get id token from Google service account credentials." + e);
```
mgyucht, line 59: This should be `LOG.error` since we cannot make progress and the application will halt.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+      }
+      Map<String, String> headers = new HashMap<>();
+      headers.put("Authorization", String.format("Bearer %s", idToken.getTokenValue()));
+
+      if (config.isAccountClient()) {
+        AccessToken token;
+        try {
+          token = finalServiceAccountCredentials.createScoped(GcpScopes).refreshAccessToken();
+        } catch (Exception e) {
+          LOG.warn("Failed to refresh access token from Google service account credentials." + e);
```
mgyucht, line 70: Ditto, `LOG.error` here.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
```diff
@@ -0,0 +1,79 @@
...
+          throw new RuntimeException(e);
+        }
+        headers.put("X-Databricks-GCP-SA-Access-Token", token.getTokenValue());
```
mgyucht, line 73: Maybe we define this constant in the GoogleUtils?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleIdCredentialsProvider.java
```diff
@@ -0,0 +1,82 @@
...
+    ImpersonatedCredentials impersonatedCredentials =
+        ImpersonatedCredentials.create(
+            googleCredentials, googleServiceAccount, null, new ArrayList<>(), 3600);
+
+    IdTokenCredentials idTokenCredentials =
+        IdTokenCredentials.newBuilder()
+            .setIdTokenProvider(impersonatedCredentials)
+            .setTargetAudience(host)
+            // Setting this will include email in the id token.
+            .setOptions(Collections.singletonList(IdTokenProvider.Option.INCLUDE_EMAIL))
```
mgyucht, line 50: Is this not needed as in google-credentials flow?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleIdCredentialsProvider.java
```diff
@@ -0,0 +1,82 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(GoogleIdCredentialsProvider.class);
+
+  @Override
+  public String authType() {
+    return "google-id";
+  }
+
+  @Override
+  public HeaderFactory configure(DatabricksConfig config) {
```
mgyucht, line 24: The same comments from the previous file apply here as well.

mgyucht (commented):
Few comments, but otherwise this is great! I would look into abstracting some of the logic about fetching tokens into GoogleUtils so you can share some implementations between these two classes. Also, we need to check that we can successfully fetch a token in the first place before returning the closure; otherwise, the SDK won't fall back to a different credential provider in case of misconfiguration.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
edwardfeng-db: As discussed offline, we don't have to do this to stay consistent with the other sdk libraries

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
```diff
@@ -29,7 +29,9 @@ public DefaultCredentialsProvider() {
             new AzureCliCredentialsProvider(),
             new ExternalBrowserCredentialsProvider(),
             new DatabricksCliCredentialsProvider(),
-            new NotebookNativeCredentialsProvider());
+            new NotebookNativeCredentialsProvider(),
+            new GoogleCredentialsCredentialsProvider(),
+            new GoogleIdCredentialsProvider());
```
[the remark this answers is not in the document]
edwardfeng-db: Yep, checked, same order

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleCredentialsCredentialsProvider.java
edwardfeng-db: Yes `IOException` is the one documented in the library code. So this should be fine

databricks-sdk-java/src/main/java/com/databricks/sdk/core/GoogleIdCredentialsProvider.java
edwardfeng-db: Right, as discussed offline, not needed in google-credentials

databricks-sdk-java/src/test/java/com/databricks/sdk/integration/CredentialsIT.java
```diff
@@ -15,10 +15,13 @@
 public class CredentialsIT {
   @Test
   void lists(AccountClient a) {
-    Iterable<Credential> list = a.credentials().list();
+    // Skipping this test for gcp because there is a permission issue with listing credentials.
```
mgyucht, line 18: Why is this?

mgyucht (approved):
What permissions do we need to change in our setup to fix CredentialsIT in GCP?

databricks-sdk-java/src/test/java/com/databricks/sdk/integration/CredentialsIT.java
edwardfeng-db: Updated the comment, as discussed offline, this API is not enabled for GCP, so skipping it

PR #202, opened 2023-12-18
Make modifications to make Azure connection work

samikshya-db:
## Changes
- Currently, U2M flow with Azure is not working, hence proposing these changes. 
- The redirection occurs to the location URL, hence, the `location` parameter is not present [while requesting for OidcEndpoints](https://github.com/databricks/databricks-sdk-java/blob/2e4946790b810a0f46ce465e7f13e085b0a897ab/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java#L531). This is fixed by disabling redirection. 
- [Discussion link](https://databricks.slack.com/archives/C01KSAWFXG8/p1702583577079619)

## Tests
Tested changes, successfully launching the browser.

Screenshot 2023-12-18 at 10 46 24 AM


databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
```diff
@@ -52,6 +52,7 @@ private RequestConfig makeRequestConfig() {
   private CloseableHttpClient makeClosableHttpClient() {
     return HttpClientBuilder.create()
         .setConnectionManager(connectionManager)
+        .disableRedirectHandling()
```
gopalldb, line 55: Is this an Azure specific change? This will disable redirect handling for all cases. @mgyucht FYI

samikshya-db: Although the test cases aren't failing atleast, but let's wait for @mgyucht to confirm.

mgyucht: Unit test cases probably won't cover this pathway. We can trigger integration tests here as well. However, I would prefer an option where you can enable/disable this on a per-request basis. If you look at the python SDK:
```python
        if self.is_azure:
            # Retrieve authorize endpoint to retrieve token endpoint after
            res = requests.get(f'{self.host}/oidc/oauth2/v2.0/authorize', allow_redirects=False)
            real_auth_url = res.headers.get('location')
            if not real_auth_url:
                return None
            return OidcEndpoints(authorization_endpoint=real_auth_url,
                                 token_endpoint=real_auth_url.replace('/authorize', '/token'))
```
We specifically disallow redirects for this one method.

HttpClient in apache commons supports a HttpRequestContext, which has a config controlling max # of redirects:
```java
  public CloseableHttpResponse execute(
            final HttpRoute route,
            final HttpRequestWrapper request,
            final HttpClientContext context,
            final HttpExecutionAware execAware) throws IOException, HttpException {
...
        final RequestConfig config = context.getRequestConfig();
        final int maxRedirects = config.getMaxRedirects() > 0 ? config.getMaxRedirects() : 50;
```
If we can plumb this through our Request interface, I think that would be best.

samikshya-db: Thanks for the idea, Miles. I have made the change

databricks-sdk-java/src/main/java/com/databricks/sdk/core/http/Request.java
```diff
@@ -74,6 +75,11 @@ public Request withHeader(String key, String value) {
     return this;
   }
 
+  public Request withParam(String key, Object value) {
```
mgyucht, line 78: Let's not couple this interface to the Apache client if possible. Can we just have an `Optional<Boolean> parameter to configure whether the request should follow redirects or not? We want to have flexibility to choose a different client in the future (e.g. when we drop java 8 support, we can use Java's built-in HTTP client).

mgyucht (commented):
Thanks for the contribution! This is much more in the right direction. Small suggestion on the interface for the request, but this approach is better.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/http/Request.java
samikshya-db: Thanks for the suggestion, @mgyucht - I added the change now - does it look ok to you?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/http/Request.java
```diff
@@ -16,6 +16,7 @@ public class Request {
   public static final String PATCH = "PATCH";
   private final String method;
   private String url;
+  private boolean redirectDisabled;
```
gopalldb, line 19: minor comment: We can also take note of if client has explicitly provided this value.

private Optional<Boolean> redirectBehavior = Optional.empty();

setRedirectBehavior(boolean redirectBehaviorValue) {
  this.redirectBehavior = Optional.of(redirectBehaviorValue);
}

Optional<Boolean> getRedirectBehavior() {
  return redirectBehavior;
}

in CommonsHttpClient:

if (!in.getRedirectBehavior().isEmpty()) {
  request.getParams().setParameter("http.protocol.handle-redirects", in.getRedirectBehavior().get());
}

samikshya-db: Sure, fixed.

mgyucht (approved):
This is perfect, thank you!

samikshya-db:
> As part of enabling in-house OAuth in Azure, OAuth team is going to change the authorize endpoint to require authentication. This means that we'll instead start redirecting to the workspace login page. We will disable authentication for /authorize without any params to ensure logic like this continues to work, but we want all tools to move away from this pattern since it is assuming that Azure workspace authorize endpoint is redirecting to some other valid endpoint URL, which is an assumption about the implementation details that is subject to change. It would be better to instead query the well known openid configuration endpoint, following redirects for it, and extract the authorize and token endpoint URLs from the JSON response.
> 
> Feel free to reach out & we can discuss more / ask questions, but I'd propose that we not move forward with this change and use the well known endpoint instead. Please include me on any reviews for such changes so I can be sure they'll continue to work with future changes to support in house oauth.

Thanks for your comment, Bryan. As per our slack discussion, we can go ahead with the changes in the PR and receive the document on updated Azure OAuth once it is ready - at a later point in time.

gopalldb (approved):
Thanks for diligently addressing comments. Wait for @miles' review as well.

mgyucht (approved):

PR #207, opened 2024-01-10
Update SDK to OpenAPI spec with bug fix

tanmay-db:
## Changes

There is a bug fix in latest OpenAPI specification. Updating SDK to use that.


databricks-sdk-java/src/main/java/com/databricks/sdk/service/vectorsearch/QueryVectorIndexRequest.java
```diff
@@ -40,7 +40,7 @@ public class QueryVectorIndexRequest {
    * vectors.
    */
   @JsonProperty("query_vector")
-  private Collection<Object /* MISSING TYPE */> queryVector;
+  private Collection<Double> queryVector;
```
tanmay-db, line 43: Highlight

mgyucht (approved):

PR #209, opened 2024-01-18
Fix Azure Databricks OAuth M2M

fjakobs:
## Changes
Fix Azure Databricks OAuth M2M

## Tests
* Manually tested with an Azure workspace
* https://github.com/databricks/eng-dev-ecosystem/actions/runs/7568909088


mgyucht (approved):
Phew, glad it was not much work!

PR #214, opened 2024-01-29
Tolerate missing GCP dependencies

mgyucht:

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
@@ -3,7 +3,6 @@ package com.databricks.sdk.core;
 import com.databricks.sdk.core.oauth.AzureServicePrincipalCredentialsProvider;
 import com.databricks.sdk.core.oauth.ExternalBrowserCredentialsProvider;
 import com.databricks.sdk.core.oauth.OAuthM2MServicePrincipalCredentialsProvider;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


edwardfeng-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
```diff
@@ -20,18 +34,20 @@ public String authType() {
...
-            new ExternalBrowserCredentialsProvider(),
-            new DatabricksCliCredentialsProvider(),
-            new NotebookNativeCredentialsProvider(),
-            new GoogleCredentialsCredentialsProvider(),
-            new GoogleIdCredentialsProvider());
+    providers = new ArrayList<>();
+    for (Class<?> clazz : providerClasses) {
+      try {
+        providers.add((CredentialsProvider) clazz.newInstance());
+      } catch (NoClassDefFoundError | InstantiationException | IllegalAccessException e) {
```
nija-at, line 41: The title says "tolerate missing gcp dependencies" but actually this PR tolerating any missing dependencies. Feels a bit like a carpet bomb, no?

What's the user experience when a user tries to use a specific provider but it fails because of a class load issue? Does it produce an error? Does the error make sense?

If the expectation is that the user will a warn, then this assumes (a) they have slf4j configured and (b) that they actually pay attention to the log message. Large enterprise applications tend to produce a lot of warns, and the user is accustomed to ignoring them.

mgyucht: You're right, this does tolerate any missing dependencies, but only those that are only depended on for authentication. If you removed dependencies that are needed for other parts of the SDK, then the SDK would crash. This is essentially saying that we'll do best effort to instantiate `CredentialsProvider` classes, but won't fail entirely if we can't. Since this only applies to the auth components, and it requires people to explicitly exclude such a library, we don't expect this to happen often at all.

A user would see the following message:
```
09:25 [WARN ] Failed to instantiate credentials provider: com.databricks.sdk.core.GoogleIdCredentialsProvider, skipping. Cause: java.lang.NoClassDefFoundError: com/google/auth/oauth2/IdTokenProvider
```

We really only have logging as a way to report non-fatal information to users today. The main alternative would be to throw an exception and ask the customer to rewrite their code to explicitly ignore this CredentialProvider, but that may be unnecessary complexity for a user.

PR #216, opened 2024-01-30
Derive config file name from attribute name

fjakobs:
## Changes
Derive config file name from attribute name

This also renames 
- `getAzureMSI` -> `getAzureMsi`
- `setAzureMSI` -> `setAzureMsi`

The old methods are kept in place but marked as deprecated.


mgyucht (approved):
Very nice! Thanks for doing this!

PR #217, opened 2024-02-01
Fix Databricks OAuth U2M on GCP

fjakobs:
## Tests

Manually tested with a profile that uses `databricks-cli` with a GCP workspace


tanmay-db (approved):
Thanks for the fix. Just to confirm, this would need to be done in other SDKs as well right? Example: https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/credentials_provider.py#L519

````
@credentials_provider('databricks-cli', ['host', 'is_aws'])
def databricks_cli(cfg: 'Config') -> Optional[HeaderFactory]:
````
Curious -- do you know why we only supported aws before?

tanmay-db:
Okay, I just saw you have another PR: https://github.com/databricks/databricks-sdk-py/pull/526 for it.

fjakobs:
Go already has this change and for JS and Python I have 

https://github.com/databricks/databricks-sdk-js/pull/134
https://github.com/databricks/databricks-sdk-py/pull/526

PR #218, opened 2024-02-01
Make new workspace client from account client

mgyucht:
## Changes
Ported to the Java SDK from https://github.com/databricks/databricks-sdk-go/pull/792/files and https://github.com/databricks/databricks-sdk-go/pull/700.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
@@ -5,13 +5,14 @@ import com.databricks.sdk.core.http.HttpClient;
 import com.databricks.sdk.core.http.Request;
 import com.databricks.sdk.core.http.Response;
 import com.databricks.sdk.core.oauth.OpenIDConnectEndpoints;
+import com.databricks.sdk.core.utils.Cloud;
 import com.databricks.sdk.core.utils.Environment;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.File;
 import java.io.IOException;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.lang.reflect.Field;
+import java.util.*;
+
 import org.apache.http.HttpMessage;
 
 public class DatabricksConfig {
@@ -123,6 +124,8 @@ public class DatabricksConfig {
   private HttpClient httpClient;
 
   private Environment env;
+
+  private DatabricksEnvironment databricksEnvironment;
 
   public Environment getEnv() {
     return env;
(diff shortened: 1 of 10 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/AzureEnvironment.java
```diff
@@ -53,28 +53,28 @@ public String getActiveDirectoryEndpoint() {
     ENVIRONMENTS.put(
         "PUBLIC",
         new AzureEnvironment(
-            "AzurePublicCloud",
+            "PUBLIC",
```
mgyucht, line 56: These names are not currently referred to anywhere.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -83,9 +83,6 @@ public class DatabricksConfig {
   @ConfigAttribute(env = "ARM_ENVIRONMENT")
   private String azureEnvironment;
 
-  @ConfigAttribute(env = "DATABRICKS_AZURE_LOGIN_APP_ID", auth = "azure")
```
mgyucht, line 86: This is now derived from the DatabricksEnvironment.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/AzureServicePrincipalCredentialsProvider.java
```diff
@@ -26,17 +26,45 @@ public HeaderFactory configure(DatabricksConfig config) {
...
       headers.put("Authorization", "Bearer " + inner.getToken().getAccessToken());
-      addWorkspaceResourceId(config, headers);
-      addSpManagementToken(cloud, headers);
+      AzureUtils.addWorkspaceResourceId(config, headers);
+      AzureUtils.addSpManagementToken(cloud, headers);
       return headers;
     };
   }
+
+  /**
```
mgyucht, line 44: Moved from AzureUtils. This logic only applies to Azure SP auth.

databricks-sdk-java/src/test/java/com/databricks/sdk/core/oauth/AzureServicePrincipalCredentialsProviderTest.java
```diff
@@ -1,60 +1,48 @@
...
+import java.util.Map;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 
 class AzureServicePrincipalCredentialsProviderTest {
   private static final String TOKEN = "t-123";
   private static final String TOKEN_TYPE = "token-type";
-  public static final String PUBLIC_MANAGEMENT_ENDPOINT = "https://management.core.windows.net/";
-
-  private static RefreshableTokenSource mockTokenSource() {
```
mgyucht, line 20: Now we mock out the HTTP Client response instead of the tokenSourceFor method.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
hectorcast-db: Does this impact existing users which are using the env var?

hectorcast-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
mgyucht: FYI: this is only dev/staging users. All prod users use the same login ID. It shouldn't matter because there is only one correct login app ID to use for your workspace, and it is coupled to the environment you're in.

PR #223, opened 2024-02-06
Only include request body when the OpenAPI operation indicates that there is a request body

mgyucht:
## Changes
When a request doesn't have a request body, even if it is a PUT or POST request, we should not serialize it in the body. See https://github.com/databricks/databricks-sdk-go/blob/main/.codegen/impl.go.tmpl#L35 for the corresponding condition in the Go SDK.

## Tests
Nightly tests pass. There is one failing nightly test right now (getWorkspaceClient) that is already failing on `main` and that I'm working on fixing.


tanmay-db (approved):
LGTM

PR #242, opened 2024-02-27
[PECO-1542] Add support for proxy

vikrantpuppala:
## Changes

This PR adds a way to provide proxy configs in the SDK directly, ability to pick up the system properties and authentication for proxy via basic or negotiate-kerberos schemes

## Tests

These changes were tested in an environment consisting of a proxy server with kerberos authentication hosted on Ubuntu and Windows VMs. The detailed testing details are documented here: https://docs.google.com/document/d/1Zxlfx-R_JytFqMZfQMBfSQ8mj4W-aHa60vw4KdjUkEc/edit


mgyucht:
@vikrantpuppala @gopalldb can you see if the linked commit works for you? We may want to consider a standard set of environment variables as well, but different languages seem to have different standards for propagating proxy information. If system properties is sufficient, I would like to stick with that for now.

vikrantpuppala:
@mgyucht I think system properties should be sufficient whenever we want to inherit system proxy. But we also need a way to provide proxy properties directly (had added those via DatabricksConfig for now in this PR to quickly test for a POC)

mgyucht:
Which properties do you need to set? I think you can set proxy username/password with `https.proxyUser/https.proxyPassword`, according to `SystemDefaultCredentialsProvider`. See https://stackoverflow.com/questions/1626549/authenticated-http-proxy-with-java.

vikrantpuppala:
In certain cases, customers would not have a system wide proxy but would have a proxy only for the jdbc driver (or different proxies across different applications). Hence, we support providing proxy host/port/creds directly to the driver which we want to pass on to the http client

databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
```diff
@@ -57,6 +76,99 @@ private CloseableHttpClient makeClosableHttpClient() {
...
+    }
+  }
+
+  public static void setupNegotiateProxyAuth(HttpClientBuilder builder, AuthScope authScope) {
+    // We only support kerberos for negotiate as of now
+    System.setProperty("javax.security.auth.useSubjectCredsOnly", "false");
+    // "java.security.krb5.conf" system property needs to be set if krb5.conf is not in the default
+    // location
+    // Use "sun.security.krb5.debug" and "sun.security.jgss.debug" system properties for debugging
+    Credentials use_jaas_creds =
```
gopalldb, line 141: nit: use consistent naming for variable: useJaasCreds?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
```diff
@@ -57,6 +76,99 @@ private CloseableHttpClient makeClosableHttpClient() {
...
+            return null;
+          }
+
+          public Principal getUserPrincipal() {
+            return null;
+          }
+        };
+
+    CredentialsProvider credsProvider = new BasicCredentialsProvider();
+    credsProvider.setCredentials(authScope, use_jaas_creds);
```
gopalldb, line 153: but we are only returning null, do we still need to set this credential provider?

vikrantpuppala: Drawing an analogy from curl: https://curl.se/docs/manpage.html#-u
>If you use a Windows SSPI-enabled curl binary and perform Kerberos V5, Negotiate, NTLM or Digest authentication then you can tell curl to select the username and password from your environment by specifying a single colon with this option: "-u :".
I believe when both username and password are not provided is when it looks at the environment to pick up the credentials

databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
```diff
@@ -57,6 +76,99 @@ private CloseableHttpClient makeClosableHttpClient() {
...
+  public static void setupProxy(ProxyConfig config, HttpClientBuilder builder) {
+    String proxyHost = null;
+    Integer proxyPort = null;
+    String proxyUser = null;
+    String proxyPassword = null;
+    if (config.isUseSystemProperties()) {
+      builder.useSystemProperties();
+      String protocol = System.getProperty("https.proxyHost") != null ? "https" : "http";
+      proxyHost = System.getProperty(protocol + ".proxyHost");
+      proxyPort = Integer.parseInt(System.getProperty(protocol + ".proxyPort"));
```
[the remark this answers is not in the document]
vikrantpuppala: If the client has set UseSystemProperties to true, they must ideally have these system properties set, otherwise these would fail with NPE when we try to parseInt from null/build an AuthScope with the relevant error i.e. xx is invalid/null

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ProxyConfig.java
```diff
@@ -0,0 +1,72 @@
...
+  private boolean useSystemProperties;
+
+  public enum ProxyAuthType {
+    // Currently we only support BASIC and SPNEGO
+    NONE,
+    BASIC,
+    // We only support kerberos for negotiate
+    SPNEGO
+  }
+
```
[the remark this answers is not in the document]
vikrantpuppala: Have modelled this similar to the DatabricksConfig where we create an empty config and set variables as needed.

gopalldb (commented):
LG, can we tests?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
```diff
@@ -57,6 +76,99 @@ private CloseableHttpClient makeClosableHttpClient() {
...
+  private CloseableHttpClient makeClosableHttpClient(ProxyConfig proxyConfig) {
+    HttpClientBuilder builder =
+        HttpClientBuilder.create()
+            .setConnectionManager(connectionManager)
+            .setDefaultRequestConfig(makeRequestConfig());
+    setupProxy(proxyConfig, builder);
+    return builder.build();
+  }
+
+  public static void setupProxy(ProxyConfig config, HttpClientBuilder builder) {
```
mgyucht, line 88: Could we move all of this proxy configuration logic into a separate file?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ProxyConfig.java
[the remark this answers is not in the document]
mgyucht: Can we move these directly into DatabricksConfig as well? I think there are use-cases of configuring different proxy endpoints for different profiles, so we will want to actually allow users to configure this in their .databrickscfg. Doing this, we may want to rename these fields to proxyHost, proxyPort, etc.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ProxyConfig.java
```diff
@@ -0,0 +1,72 @@
+package com.databricks.sdk.core;
+
+public class ProxyConfig {
+  private String host;
+  private int port;
+  private String username;
+  private String password;
+  private ProxyAuthType proxyAuthType;
+  private boolean useSystemProperties;
```
mgyucht, line 9: Should this default to `true`? If users want to use system properties, it seems strange that they would need to set it explicitly.

If we're introducing this, we should also use it to control whether `.useSystemProperties()` is called on line 75.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ProxyConfig.java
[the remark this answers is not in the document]
vikrantpuppala: Do we want to have these properties in both DatabricksConfig as well as ProxyConfig? I've added it to both for now, but can remove ProxyConfig if you think that's unnecessary

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ProxyConfig.java
vikrantpuppala: Not sure about this one, I felt like my app should be independent of system proxy settings unless i choose that explicitly but I understand the other way could also feel natural.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -114,6 +114,24 @@ public class DatabricksConfig {
...
+  private Integer proxyPort;
+
+  @ConfigAttribute(env = "PROXY_USERNAME")
+  private String proxyUsername;
+
+  @ConfigAttribute(env = "PROXY_PASSWORD")
+  private String proxyPassword;
+
+  @ConfigAttribute(env = "PROXY_AUTH_TYPE")
+  private ProxyConfig.ProxyAuthType proxyAuthType;
```
mgyucht, line 130: How does this work if someone puts it in their `.databrickscfg`? You might need to add logic to `ConfigAttributeAccessor.setValueOnConfig`

databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
```diff
@@ -41,6 +44,23 @@ public CommonsHttpClient(int timeoutSeconds) {
...
 
+  public CommonsHttpClient(DatabricksConfig databricksConfig) {
+    int timeoutSeconds =
+        databricksConfig.getHttpTimeoutSeconds() == null
+            ? 300
+            : databricksConfig.getHttpTimeoutSeconds();
+    timeout = timeoutSeconds * 1000;
+    connectionManager.setMaxTotal(100);
+    ProxyConfig proxyConfig = new ProxyConfig(databricksConfig);
+    hc = makeClosableHttpClient(proxyConfig);
```
mgyucht, line 55: Can we condense the logic into a single constructor?
```suggestion
    this(databricksConfig.getHttpTimeoutSeconds() == null
            ? 300
            : databricksConfig.getHttpTimeoutSeconds(),
            new ProxyConfig(databricksConfig));
```

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ProxyConfig.java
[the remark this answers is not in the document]
mgyucht: I think it's fine to have a separate config that is smaller for just the proxy elements, as you've done.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -462,6 +476,60 @@ public DatabricksConfig setHttpClient(HttpClient httpClient) {
...
+  public ProxyConfig.ProxyAuthType getProxyAuthType() {
+    return proxyAuthType;
+  }
+
+  public DatabricksConfig setProxyAuthType(ProxyConfig.ProxyAuthType proxyAuthType) {
+    this.proxyAuthType = proxyAuthType;
+    return this;
+  }
+
+  public Boolean isUseSystemPropertiesHttp() {
```
mgyucht, line 524: nit: `getUseSystemPropertiesHttp`

mgyucht (commented):
Mostly LGTM, few smaller remarks.

mgyucht (approved):

PR #247, opened 2024-03-15
Lock around field accessibility changes

webmakersteve:
Adjusts the field accessibility changes so they are protected behind a lock. Epic was running into this when we concurrently created WorkspaceClients and they would often read properties that another thread had re-set to unaccessible.

## Changes
Synchronize on the `field` object before we mutate and read it to ensure accessibility maintains throughout the code block

## Tests
No new tests. Nothing new added so existing tests should ensure logic is correct for config reading. Testing multithreaded races in this case would prove difficult.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
@@ -349,19 +349,24 @@ public class ApiClient {
       if (firstHeader == null) {
         continue;
       }
-      try {
-        field.setAccessible(true);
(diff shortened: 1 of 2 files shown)


mgyucht:
@webmakersteve Thanks for making this PR! We require all commits to be signed in this repo. Can you please setup git commit signing following the instructions at https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits, sign this commit & force push?

mgyucht (approved):
LGTM! Thanks for this!

webmakersteve:
Signed the commit!

PR #249, opened 2024-03-19
[PECO-1008] Add logic for retries

vikrantpuppala:
## Changes

https://docs.google.com/document/d/1GobYafEUAIHqXjLF8n7FuDB3_BxDnEZkEc_CQglR8O0/edit
Added changes to retry requests only if they are idempotent

## Tests

Unit tests


databricks-sdk-java/src/main/java/com/databricks/sdk/core/retry/StatementExecutionRetryStrategyPicker.java
```diff
@@ -0,0 +1,31 @@
+package com.databricks.sdk.core.retry;
+
+import com.databricks.sdk.core.http.Request;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class StatementExecutionRetryStrategyPicker implements RetryStrategyPicker {
```
gopalldb, line 8: Can we have better name for this class? It has nothing to do with StatementExecution. Something like ListBasedRetryStrategyPicker or RetryStrategyPickerImpl etc

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -278,14 +285,39 @@ private Response executeInner(Request in) {
...
+
+  private static Optional<Long> getBackoffFromRetryAfterHeader(Response response) {
+    List<String> retryAfterHeader = response.getHeaders("retry-after");
+    if (retryAfterHeader.isEmpty()) {
+      return Optional.empty();
+    }
+    long waitTime = 0;
+    for (String retryAfter : retryAfterHeader) {
+      try {
+        Date date = Date.from(Instant.parse(retryAfter));
```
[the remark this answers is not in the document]
vikrantpuppala: Could be either: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After (hence, the try-catch)

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -278,14 +285,39 @@ private Response executeInner(Request in) {
...
 
-  private int getBackoffMillis(int attemptNumber) {
-    int maxWait = 10000;
+  private long getBackoffMillis(Response response, int attemptNumber) {
+    Optional<Long> backoffMillisInResponse = getBackoffFromRetryAfterHeader(response);
+    if (backoffMillisInResponse.isPresent()) {
+      return backoffMillisInResponse.get();
+    }
+    int minWait = 1000; // 1 second
+    int maxWait = 60000; // 1 minute
```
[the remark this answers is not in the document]
vikrantpuppala: https://docs.google.com/document/d/1NGAYZ2JlsC57gCjw6-nJ8QAPGBplUzdD7XhGR2RjLRA/edit
For dbsql, the original recommendation was 30s, but Swee had recommended to bring it up to 60s

databricks-sdk-java/pom.xml
```diff
@@ -93,16 +93,4 @@
       <version>1.20.0</version>
     </dependency>
   </dependencies>
-  <build>
-    <plugins>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-compiler-plugin</artifactId>
-        <configuration>
```
gopalldb: why are we removing this?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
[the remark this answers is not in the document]
gopalldb: Do we want to keep this configurable?

gopalldb (commented):
LG overall, few minor comments

databricks-sdk-java/pom.xml
vikrantpuppala: I had added this, but caused failures, so removed

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -255,37 +263,62 @@ private Response executeInner(Request in) {
       // 2. The request failed with a non-retriable error (err != null, out == null).
       // 3. The request failed with a retriable error, but the number of attempts exceeds
       // maxAttempts.
-      DatabricksError res = ApiErrors.checkForRetry(out, err);
-      if (!res.isRetriable()) {
-        if (res.getErrorCode() == null) {
+      if (!retryStrategy.isRetriable(out, err)) {
+        if (retryStrategy.getError() == null || retryStrategy.getError().getErrorCode() == null) {
```
mgyucht, line 267: It feels kind of wrong to me that the retry strategy is returning the error, and it also results in internal mutability which I would like to avoid if possible.

To do this, I think we need to separate the two independent retry conditions, one based on the exception thrown by the HTTP client, the other based on the error response from Databricks.

Algorithm:
1. If exception is not null, check the exception. Retry if retriable, otherwise don't.
2. No exception. It's an error response that we can read using `ApiErrors.readErrorFromResponse`.
3. Then, check the error response for retry conditions.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -255,37 +263,62 @@ private Response executeInner(Request in) {
...
+    long waitTime = 0;
+    for (String retryAfter : retryAfterHeader) {
+      try {
+        Date date = Date.from(Instant.parse(retryAfter));
+        LocalDateTime dateTime = LocalDateTime.ofInstant(date.toInstant(), ZoneId.systemDefault());
+        LocalDateTime now = LocalDateTime.now();
+        waitTime = java.time.Duration.between(now, dateTime).getSeconds();
+      } catch (Exception e) {
+        // If not a date, assume it is seconds
+        waitTime = Long.parseLong(retryAfter);
```
mgyucht, line 318: If this throws, let's just return `Optional.empty()`.

databricks-sdk-java/src/test/java/com/databricks/sdk/core/ApiClientTest.java
```diff
@@ -100,6 +123,16 @@ private SuccessfulResponse getTooManyRequestsResponse(Request req) {
         new Response(req, 429, "Too Many Requests", Collections.emptyMap(), (String) null));
   }
 
+  private SuccessfulResponse getTooManyRequestsResponseWithRetryAfterHeader(Request req) {
+    return new SuccessfulResponse(
+        new Response(
+            req,
+            429,
+            "Too Many Requests",
+            Collections.singletonMap("retry-after", Collections.singletonList("1")),
```
mgyucht, line 132: Can we add a case for retry-after with absolute date as well, if we're going to support it?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/retry/RetryStrategy.java
```diff
@@ -0,0 +1,10 @@
+package com.databricks.sdk.core.retry;
+
+import com.databricks.sdk.core.DatabricksError;
+import com.databricks.sdk.core.http.Response;
+
+public interface RetryStrategy {
```
mgyucht, line 6: Please add doccomments to this.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/retry/RetryStrategyPicker.java
```diff
@@ -0,0 +1,7 @@
+package com.databricks.sdk.core.retry;
+
+import com.databricks.sdk.core.http.Request;
+
+public interface RetryStrategyPicker {
```
mgyucht, line 5: Ditto: doccomments

databricks-sdk-java/src/main/java/com/databricks/sdk/core/retry/IdempotentRequestRetryStrategy.java
```diff
@@ -0,0 +1,17 @@
+package com.databricks.sdk.core.retry;
+
+import com.databricks.sdk.core.DatabricksError;
+import com.databricks.sdk.core.http.Response;
+
+public class IdempotentRequestRetryStrategy implements RetryStrategy {
+  @Override
+  public boolean isRetriable(Response response, Exception e) {
+    // TODO: Not all errors are retriable, we can skip retrying in certain cases
```
mgyucht, line 9: This is definitely not right. We hit this codepath when actual 4xx errors are returned that mustn't be retried in the client (e.g. Bad Request, Resource Not Found, etc.). 

I think there are 4 classes of errors here:
1. The server definitely rejected the request with a non-retriable error.
2. The server possibly received the request, but there was a timeout, either client-side or server-side, that prevented the request from being handled.
3. The server rejected the request with a retriable error. Today, this includes 429, 501 and 503 and "transient errors" below.
4. The server definitely did not receive the request.

I think IdempotentRetryStrategy retries on 2, 3, and 4, and NonIdempotentRetryStrategy retries on 3 and 4 only.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/retry/RequestBasedRetryStrategyPicker.java
```diff
@@ -0,0 +1,42 @@
+package com.databricks.sdk.core.retry;
+
+import com.databricks.sdk.core.DatabricksConfig;
+import com.databricks.sdk.core.http.Request;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class RequestBasedRetryStrategyPicker implements RetryStrategyPicker {
+  private static final List<Request> NON_IDEMPOTENT_REQUESTS =
```
mgyucht, line 10: I thought we were going to do this the other way around: treat everything as non-idempotent but have an allow-list for the things which are allowed to retry even when we're not sure the server has processed the request.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/retry/RequestBasedRetryStrategyPicker.java
```diff
@@ -0,0 +1,42 @@
...
+    this.nonIdempotentRequestsWithUrl =
+        NON_IDEMPOTENT_REQUESTS.stream()
+            .map(request -> new Request(request.getMethod(), config.getHost() + request.getUrl()))
+            .collect(Collectors.toList());
+  }
+
+  @Override
+  public RetryStrategy getRetryStrategy(Request request) {
+    if (isNonIdempotentRequest(request)) {
+      return new NonIdempotentRequestRetryStrategy();
```
mgyucht, line 27: Can we avoid allocation and just use a static object if possible?

mgyucht (commented):
First batch of comments. Approach seems alright, but a few things need to be worked out.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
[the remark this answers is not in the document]
vikrantpuppala: I believe we can instead provide an overall timeout limit that is configurable i.e. wait time across all retries

databricks-sdk-java/src/main/java/com/databricks/sdk/core/retry/RequestBasedRetryStrategyPicker.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class RequestBasedRetryStrategyPicker implements RetryStrategyPicker {
+  private static final List<Request> NON_IDEMPOTENT_REQUESTS =
+      Arrays.asList(new Request("POST", "/api/2.0/sql/statements/"));
+
+  private final List<Request> nonIdempotentRequestsWithUrl;
+  DatabricksConfig config;
```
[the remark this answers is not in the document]
vikrantpuppala: sorry, what is the issue here?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/retry/IdempotentRequestRetryStrategy.java
vikrantpuppala: I added it to IdempotentRequestRetryStrategy as a do not retry if 1 which is opposite to our approach for NonIdempotentRequestRetryStrategy i.e. retry on 3, 4. It seems a bit counter-intuitive at a high level but seems like the right thing to do? Let me know your thoughts

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -63,11 +70,12 @@ public ApiClient(DatabricksConfig config, Timer timer) {
       debugTruncateBytes = 96;
     }
 
-    maxAttempts = 3;
+    maxAttempts = 4;
```
mgyucht, line 73: Just curious: our other SDKs implement retries based on timeouts: retry until time has elapsed, then fail. Would it be problematic for your use case? Either way, let's not make that change in this PR to reduce complexity.

mgyucht (commented):
Generally looking good. A couple smaller comments, but otherwise I think the approach is sound.

Can you add a test for the retry with backoff checking that the time waited for each retry grows (roughly) exponentially?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -249,43 +258,76 @@ private Response executeInner(Request in) {
...
             String.format("Request %s failed after %d retries", in, maxAttempts), err);
       }
 
       // Retry after a backoff.
-      int sleepMillis = getBackoffMillis(attemptNumber);
+      long sleepMillis = getBackoffMillis(out, attemptNumber);
       LOG.debug(String.format("Retry %s in %dms", in.getRequestLine(), sleepMillis));
       try {
-        timer.wait(sleepMillis);
+        Thread.sleep(sleepMillis);
```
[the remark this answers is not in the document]
vikrantpuppala: The tests fail with timer
```
java.lang.IllegalMonitorStateException
	at java.base/java.lang.Object.wait(Native Method)
	at com.databricks.sdk.core.ApiClient.executeInner(ApiClient.java:280)
	at com.databricks.sdk.core.ApiClient.getResponse(ApiClient.java:227)
	at com.databricks.sdk.core.ApiClient.execute(ApiClient.java:218)
```

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
vikrantpuppala: I think that would be a good idea here as well i.e. having an overall timeout when we stop retrying

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -276,10 +276,12 @@ private Response executeInner(Request in) {
...
-      try {
-        Thread.sleep(sleepMillis);
-      } catch (InterruptedException ex) {
-        Thread.currentThread().interrupt();
+      synchronized (timer) {
+        try {
+          timer.wait(sleepMillis);
+        } catch (InterruptedException ex) {
+          Thread.currentThread().interrupt();
+        }
```
mgyucht, line 284: This issue is probably because `wait()` is actually inherited from `Object` and is used for synchronization. Let's rename this method to `sleep()`, then we won't need this here.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/retry/RequestBasedRetryStrategyPicker.java
```diff
@@ -21,15 +21,15 @@ public class RequestBasedRetryStrategyPicker implements RetryStrategyPicker {
           // Create a new session v2.0
           new Request("POST", "/api/2.0/sql/sessions/"),
           // Delete an existing session v1.0
-          new Request("DELETE", "/api/2.0/sql/statements/sessions/*"),
+          new Request("DELETE", "/api/2.0/sql/statements/sessions/."),
```
mgyucht, line 24: Does it not need to be like this for all of these?
```suggestion
          new Request("DELETE", "/api/2.0/sql/statements/sessions/.*"),
```

mgyucht (approved):
Two small nits, otherwise LGTM.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
vikrantpuppala: Ah, my bad, I hadn't realised timer was actually defined in the sdk

vikrantpuppala:
Merged in #264

PR #255, opened 2024-04-02
fix to set `filter` to `null` when `next_page_token` is returned from previous call

kai-zhu-sonatype:
## Changes
clear `request.filter` when preivous call returns next_page_token to avoid error when with filter request
```bash
databricks.sdk.errors.platform.InvalidParameterValue: When providing a page token and a filter, the filter must match the filter represented by the page token.
```
similar fix in https://github.com/databricks/databricks-sdk-py/pull/605, https://github.com/databricks/databricks-sdk-go/pull/875
## Tests
- [x] make test run locally
- [x] make fmt applied

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
@@ -342,13 +342,17 @@ public class DatabricksConfig {
     return this;
   }
 
-  /** @deprecated Use {@link #getAzureUseMsi()} instead. */
+  /**
+   * @deprecated Use {@link #getAzureUseMsi()} instead.
+   */
   @Deprecated()
   public boolean getAzureUseMSI() {
     return azureUseMsi;
   }
 
-  /** @deprecated Use {@link #setAzureUseMsi(boolean)} instead. */
+  /**
+   * @deprecated Use {@link #setAzureUseMsi(boolean)} instead.
+   */
   @Deprecated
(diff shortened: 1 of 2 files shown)


kai-zhu-sonatype:
> Thanks for making these changes! Unfortunately, this code is autogenerated, so these changes will be overwritten in the next release.
> 
> Instead, I'm going to follow-up with the Delta Live Tables team to see if they can modify their API to tolerate this condition. I believe the same filter should be used for all requests, so this is probably a server-side issue.

Thanks @mgyucht for your response, I was checking where the gencode specification file is when running `make gen`, didn't find `.openapi-codegen.json` or a similar file, only test file in `openapi/testdata/spec.json`

mgyucht:
I understand. This could be fixed by making client-side changes, but the service behavior actually violates guidelines for paginated endpoints that we mandate for all services to simplify the maintenance of the SDKs. Fixing server-side also means that it will work for existing clients without them needing to update the SDK to the latest version. As it is a non-breaking change, I will try that avenue first.

kai-zhu-sonatype:
> I understand. This could be fixed by making client-side changes, but the service behavior actually violates guidelines for paginated endpoints that we mandate for all services to simplify the maintenance of the SDKs. Fixing server-side also means that it will work for existing clients without them needing to update the SDK to the latest version. As it is a non-breaking change, I will try that avenue first.

Thanks @mgyucht , was been told from a Databricks support
> filter should be removed from the request if page_token is specified.
"Page token returned by previous call. This field is mutually exclusive with all fields in this request except max_results. An error is returned if any fields other than max_results are set when this field is set."

if could be fixed from `server-side` that would be good option.

PR #272, opened 2024-04-23
Fix interrupt handling

avijeet95:
## Changes
Current implementation does not work well when client interrupts a thread where SDK code is running. Once interrupted, the logic continues to run till the timeout (default 20mins) without sleeping at all. Each time thread.sleep is called, it throws interrupt exception and this results in calling the status API continuously without delay. 

The PR is changing the handling to throw DatabricksException when the thread is interrupted so that clients can stop the processing and handle the exception as they wish.

## Tests
All unit tests are successful.

### DIFF
--- a/.codegen/api.java.tmpl
+++ b/.codegen/api.java.tmpl
@@ -96,6 +96,7 @@ public class {{.PascalName}}API {
             Thread.sleep((long) (sleep * 1000L + Math.random() * 1000));
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
+            throw new DatabricksException("Current thread was interrupted", e);
         }
         attempt++;
     }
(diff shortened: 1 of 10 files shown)


.codegen/api.java.tmpl
```diff
@@ -95,7 +95,7 @@ public class {{.PascalName}}API {
         try {
             Thread.sleep((long) (sleep * 1000L + Math.random() * 1000));
         } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
+            throw new DatabricksException("Current thread was interrupted", e);
```
mgyucht, line 98: You have to call `Thread.currentThread().interrupt()` to reset the interrupted bit so that callers will know that this thread was interrupted and can handle such a change.

mgyucht (commented):
Do you need to do the same in ApiClient.java?

.codegen/api.java.tmpl
avijeet95: I thought that we are already throwing an exception which should be handled by the callers. But I think it won't harm to reset the flag. Just in case someone suppresses the exception, it will still flag the thread as interrupted.

avijeet95:
> Do you need to do the same in ApiClient.java?

Thanks for catching this. I'll update it as well.

mgyucht (approved):

PR #281, opened 2024-05-06
Remove unused script from repository root

pietern:


pietern:
Doesn't seem to be used per a `git grep tmp.sh`.

PR #285, opened 2024-05-08
Fix NPE in string function on DatabricksConfig

pietern:
## Changes

If the configuration isn't resolved, the `env` field will be `null`, and the `toString` function would throw an NPE. This change fixes that and adds a few tests. I changed the string output to always be non-empty, even if no properties are set.

## Tests

Unit tests.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ConfigLoader.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ConfigLoader.java
@@ -1,5 +1,6 @@
 package com.databricks.sdk.core;
 
+import com.databricks.sdk.core.utils.Environment;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -8,8 +9,6 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.file.Paths;
 import java.util.*;
-
-import com.databricks.sdk.core.utils.Environment;
 import org.ini4j.Ini;
 import org.ini4j.Profile;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/ConfigLoader.java
```diff
@@ -197,7 +198,13 @@ public static String debugString(DatabricksConfig cfg) {
       List<String> attrsUsed = new ArrayList<>();
       List<String> buf = new ArrayList<>();
 
-      Map<String, String> getEnvAllEnv = cfg.getEnv().getEnv();
+      Environment env = cfg.getEnv();
```
tanmay-db, line 201: If my understanding is correct, then this would only happen for when we do `config.toString()` since we directly call debugString, for the other case (where we call debugString in makeNicerError) we have a check that checks for the null env:

````
    if (cfg.getEnv() != null) {
      debugString = debugString(cfg);
    }
````

Since we are putting a null check at the base (in debugString), we might have to remove this check in makeNicerError and update the if condition on line 189, since debugString would never be empty: `!debugString.isEmpty()`

Otherwise the message that we get from makeNicerError won't contain `Env: <none>`, `Config: <empty>`

tanmay-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ConfigLoader.java
mgyucht: I don't think we need to change anything about that other callsite. Basically, it's saying that we won't include the config in the `makeNicerError` message if it is not yet resolved, but we can still include that in DatabricksConfig's `toString()`

mgyucht (approved):

PR #291, opened 2024-05-17
Support partners in SDK

mgyucht:
## Changes
This PR is a port of https://github.com/databricks/databricks-sdk-go/pull/925 to the Java SDK.

Partners of Databricks need a mechanism to register themselves in libraries or applications that they write. In this way, requests made by users of those libraries will include sufficient information to link those requests to the original users.

This PR adds a new function `withPartner` that can be used by a partner to add partner information to the User-Agent header for requests made by the SDK. The new header will have the form `partner/<parther id>`. The partner identifier is opaque for the SDK, but it must be alphanumeric.

This PR also lifts the requirement that there may only be a single copy of an entry in the user agent. This allows multiple partners to register themselves in the same library or application.

Additionally, this PR adds validation on keys and values of entries in the user agent to match the Go SDK: alphanumeric keys and alphanumeric or semver values.

## Tests
Unit tests cover the implementation of the `UserAgent` class.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/UserAgent.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/UserAgent.java
@@ -1,6 +1,8 @@
 package com.databricks.sdk.core;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
@@ -10,6 +12,8 @@ public class UserAgent {
 
   private static final Map<String, String> otherInfo = new HashMap<>();
 
+  private static final List<String> partners = new ArrayList<>();
+
   // TODO: check if reading from
   // /META-INF/maven/com.databricks/databrics-sdk-java/pom.properties
   // or getClass().getPackage().getImplementationVersion() is enough.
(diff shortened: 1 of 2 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/UserAgent.java
```diff
@@ -20,8 +39,62 @@ public static void withProduct(String product, String productVersion) {
...
+  }
+
+  public static boolean matchAlphanum(String s) throws IllegalArgumentException {
+    if (regexpAlphanum.matcher(s).matches()) {
+      return true;
+    }
+    throw new IllegalArgumentException("Invalid alphanumeric string: " + s);
+  }
+
+  public static boolean matchAlphanumOrSemVer(String s) throws IllegalArgumentException {
```
tanmay-db, line 83: nit: Might be better to use above 2 methods here and throw their respective errors for clearer error message on what is invalid (alphanumeric or semver).

tanmay-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/UserAgent.java
mgyucht: You can't use the two methods above as-is; that would only work if we needed to validate that something was _both_ alphanumeric or semver.

PR #300, opened 2024-06-23
[Feature] Specify proxy auth explicitly when using system proxy

vikrantpuppala:
## Changes

In certain cases, customers can use different ways to authenticate proxy even when using the system proxy settings (basic, kerberos etc.). Hence, post this change, the auth for system proxy can be done in different ways and needs to be explicitly specified.

## Tests

Tested on a proxy setup in VMs


vikrantpuppala:
@nfx can you please take a look at this small change?

renaudhartert-db (approved):
This looks reasonable to me. 

Thinking-out-loud: Rather than having two config surfaces in ProxyUtils, I'd like to explore ways to parse the system properties higher in the call stack (e.g. when the config object is created) and use them to instantiate the config. This out of the scope of this PR though.

PR #303, opened 2024-07-03
Test parsing of error messages with `int` error codes

renaudhartert-db:
## Changes

The `/api/2.0/preview/scim/v2/Me` sometime returns the HTTP status (an `int`) as error code (which is expected to be a `string`). This is not a problem in the Java SDK because Jackson automatically converts numbers to strings. This PR makes it clear that this behavior is intended by adding a test for it.


mgyucht (approved):
Terrific, thanks!

PR #290, opened 2024-05-17
Better error message when private link enabled workspaces reject requests

mgyucht:
## Changes
Port of https://github.com/databricks/databricks-sdk-go/pull/924 to the Java SDK.

When a user tries to access a Private Link-enabled workspace configured with no public internet access from a different network than the VPC endpoint belongs to, the Private Link backend redirects the user to the login page, rather than outright rejecting the request. The login page, however, is not a JSON document and cannot be parsed by the SDK, resulting in this error message:

```
$ databricks current-user me
Error: unexpected error handling request: invalid character '<' looking for beginning of value. This is likely a bug in the Databricks SDK for Go or the underlying REST API. Please report this issue with the following debugging information to the SDK issue tracker at https://github.com/databricks/databricks-sdk-go/issues. Request log:
GET /login.html?error=private-link-validation-error:<WSID>
> * Host: 
> * Accept: application/json
> * Authorization: REDACTED
> * Referer: https://adb-<WSID>.azuredatabricks.net/api/2.0/preview/scim/v2/Me
> * User-Agent: cli/0.0.0-dev+5ed10bb8ccc1 databricks-sdk-go/0.39.0 go/1.22.2 os/darwin cmd/current-user_me auth/pat
< HTTP/2.0 200 OK
< * Cache-Control: no-cache, no-store, must-revalidate
[...]

PR #305, opened 2024-07-04
Accept any `HttpRequest` instead of `HttpUriRequest` in `CommonHttpClient`

renaudhartert-db:
## Changes

This PR addresses a regression reportedly introduced in PR #290 when connecting via a proxy that requires authentication (see context).

The problem comes from a casting exception in `CommonHttpClient` which attempts to cast a `BasicHttpRequest` (which represents the next request after receiving the auth required response) into a `HttpUriRequest`. This PR solves the casting issue by processing the ancestor of both classes, `HttpRequest`. 

### Context

The regression was detected when trying to authenticate via a Proxy that requires authentication. In particular, the request to respond to the proxy's authentication request is automatically created as a `BasicHttpRequest` by the apache `HttpClient`.   

## Tests

Unit tests and integration tests are passing. The fix was also verified by the user who uncovered the issue.

Note: we should add a regression test that simulate the proxy setting. However, I'd like to take the time to experiment with different configurations. Given that the change is relatively simple, I'd recommend to proceed with this PR and add the test in a follow-up PR.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
@@ -30,7 +30,6 @@ import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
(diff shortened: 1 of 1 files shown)


tanmay-db (approved):
LGTM. Just for my understanding, how was the regressions detected?

renaudhartert-db:
> LGTM. Just for my understanding, how was the regressions detected?

Thanks! I've added some context to the PR description

PR #307, opened 2024-07-05
Add credential provider for Azure Github OIDC

renaudhartert-db:

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/AzureGithubOidcCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/AzureGithubOidcCredentialsProvider.java
@@ -1,0 +1,100 @@
+package com.databricks.sdk.core.oauth;
+
+import com.databricks.sdk.core.*;
+import com.databricks.sdk.core.http.Request;
+import com.databricks.sdk.core.http.Response;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * {@code AzureGithubOidcCredentialsProvider} is a credentials provider for GitHub Actions that use
+ * an Azure Active Directory Federated Identity. It authenticates with Azure by exchanging GitHub's
+ * OIDC token for an Azure Active Directory (AAD) Service Principal OAuth token. This class handles
+ * the process of obtaining, refreshing, and attaching the necessary tokens to each HTTP request.
+ */
+public class AzureGithubOidcCredentialsProvider implements CredentialsProvider {
+  private final ObjectMapper mapper = new ObjectMapper();
+
+  @Override
+  public String authType() {
+    return "github-oidc-azure";
+  }
+
+  @Override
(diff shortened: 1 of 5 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/OidcTokenSource.java
```diff
@@ -0,0 +1,83 @@
...
+    } catch (IOException e) {
+      throw new DatabricksException(
+          "Failed to request auth token: corrupted token: " + e.getMessage());
+    }
+
+    if (resp.getErrorCode() != null) {
+      throw new IllegalArgumentException(resp.getErrorCode() + ": " + resp.getErrorSummary());
+    }
+    LocalDateTime expiry = LocalDateTime.now().plusSeconds(resp.getExpiresIn());
+    return new Token(resp.getAccessToken(), resp.getTokenType(), resp.getRefreshToken(), expiry);
```
tanmay-db, line 81: nit: Might be good to separate this out to another method where we parse the response.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/AzureGithubOidcCredentialsProvider.java
```diff
@@ -0,0 +1,100 @@
...
+    ObjectNode jsonResp;
+    try {
+      jsonResp = mapper.readValue(resp.getBody(), ObjectNode.class);
+    } catch (IOException e) {
+      throw new DatabricksException(
+          "Failed to request ID token: corrupted token: " + e.getMessage());
+    }
+
+    return Optional.ofNullable(jsonResp.get("value").textValue());
+  }
```
tanmay-db, line 99: same here, we could split this into another method to parse once we get the response above.

tanmay-db (approved):
LGTM

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/OidcTokenSource.java
renaudhartert-db: We could but I'm not convinced that it actually would make the code more readable given how small these parsing operations are: we really just have 2 simple steps with a `try ... catch`. 

One of the big advantages of keeping these steps at the same level is that it makes it very clear that we intent to catch (and convert) exceptions. Splitting the code would make that less clear as we would likely not add a `try ... catch` around the new `parseResponse` function:

```java
  protected Token refresh() {
    Response rawResp;
    try {
      rawResp = httpClient.execute(new FormRequest(tokenUrl, params));
    } catch (IOException e) {
      throw new DatabricksException("Failed to request auth token: " + e.getMessage(), e);
    }

    return parseToken(resp); // no need to catch here because exception or normalized in the function.
  }
```

I'd thus tend to keep it as it is except if you feel strongly against it.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/AzureGithubOidcCredentialsProvider.java
renaudhartert-db: See comment above, I'd tend to keep as is for similar reason but I'm happy to split if you feel strongly against it.

PR #319, opened 2024-07-23
[Internal] Test that Jobs API endpoints are pinned to 2.1

ricardoc-db:
## Changes

Adds regression tests to ensure required Jobs endpoints say pinned to API 2.1

## Tests

Unit tests.


ricardoc-db:
No idea why the unit test work with MacOS but are failing on ubuntu in the checks. If anyone has tips, I would appreciate it :)

dvkuznietsov (approved):

ricardoc-db:
> No idea why the unit test works with MacOS but are failing on ubuntu in the checks. If anyone has tips, I would appreciate it :)

@mgyucht Do you happen to have any idea? I see other tests using a similar strategy, and this test works on macos, so I'm really not sure what I could be missing.
If not, I'll just try re-writing it in some other way.

renaudhartert-db (approved):
LGTM

PR #324, opened 2024-07-25
Jobs GetRun 2.2 behind the scenes pagination

ricardoc-db:
## Changes

Jobs GetRun 2.2 behind the scenes pagination

## Tests

Unit tests


.codegen/_openapi_sha
```diff
@@ -1 +1 @@
-37e2bbe0cbcbbbe78a06a018d4fab06314a26a40
\ No newline at end of file
+universe:/Users/ricardo.costa/universe
```
ricardoc-db, line 1: Still looking to get an official SHA here.

databricks-sdk-java/src/main/java/com/databricks/sdk/mixin/JobsExt.java
```diff
@@ -0,0 +1,50 @@
...
+
+    Run currRun = run;
+    while (currRun.getNextPageToken() != null) {
+      request.setPageToken(currRun.getNextPageToken());
+      currRun = super.getRun(request);
+      if (paginatingIterations) {
+        run.getIterations().addAll(currRun.getIterations());
+      } else {
+        run.getTasks().addAll(currRun.getTasks());
+      }
```
dvkuznietsov, line 42: we need this condition because in case of For each run, the singular task keeps showing up in the tasks array?

databricks-sdk-java/src/main/java/com/databricks/sdk/mixin/JobsExt.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    // now that we've added all pages to the Run, the token is useless
+    run.setNextPageToken(null);
```
dvkuznietsov, line 46: previous page token too I suppose

databricks-sdk-java/src/main/java/com/databricks/sdk/mixin/JobsExt.java
```diff
@@ -0,0 +1,50 @@
...
+    super(apiClient);
+  }
+
+  public JobsExt(JobsService mock) {
+    super(mock);
+  }
+
+  @Override
+  public Run getRun(GetRunRequest request) {
+    Run run = super.getRun(request);
```
gkiko10, line 21: Is this calling API 2.2 or 2.1? I can't find where the version is defined

ricardoc-db: Currently 2.1. The implementation is supposed to work with both (as long as the types are there).

https://github.com/databricks/databricks-sdk-java/blob/c7f2582f6344ada70c345e9b09d3b128d089d102/databricks-sdk-java/src/main/java/com/databricks/sdk/service/jobs/JobsImpl.java#L98

databricks-sdk-java/src/main/java/com/databricks/sdk/mixin/JobsExt.java
ricardoc-db: My understanding is that we're only ever paginating one of the two fields, and never both at the same time.
This way we avoid having duplicate entries in the array that is not being paginated.

databricks-sdk-java/src/main/java/com/databricks/sdk/mixin/JobsExt.java
ricardoc-db: [dc03e00](https://github.com/databricks/databricks-sdk-java/pull/324/commits/dc03e0016a7159d2fa7b55e6d207a9d397c396a3)

databricks-sdk-java/src/main/java/com/databricks/sdk/mixin/JobsExt.java
gkiko10: If I understand correctly you will override the `String path` variable later?

ricardoc-db: `JobsImpl.java` is generated. It follows that the `String path` there is also generated. My understanding is that when the version is updated on the proto, the next generation will automatically update those paths.

gkiko10 (approved):
Thanks for the change

databricks-sdk-java/src/test/java/com/databricks/sdk/mixin/JobsExtTest.java
```diff
@@ -0,0 +1,85 @@
...
+    assertEquals(expectedRun, run);
+    verify(service, times(3)).getRun(any());
+  }
+
+  private void addTasks(Run run, long... taskRunIds) {
+    Collection<RunTask> tasks = new ArrayList<>();
+    for (long runId : taskRunIds) {
+      tasks.add(new RunTask().setRunId(runId));
+    }
+    run.setIterations(tasks);
```
mgyucht, line 75: You mean `run.setTasks` I assume?

databricks-sdk-java/src/main/java/com/databricks/sdk/mixin/JobsExt.java
```diff
@@ -0,0 +1,51 @@
...
+  public JobsExt(ApiClient apiClient) {
+    super(apiClient);
+  }
+
+  public JobsExt(JobsService mock) {
+    super(mock);
+  }
+
+  @Override
+  public Run getRun(GetRunRequest request) {
```
mgyucht, line 18: Ditto here, we may want to have a separate doccomment.

mgyucht (approved):
Overall LGTM. Were you able to test this out against a real workspace?

databricks-sdk-java/src/test/java/com/databricks/sdk/mixin/JobsExtTest.java
ricardoc-db: Indeed, thanks for spotting that. The test was passing previously because the expected object was also being constructed in the same way.
Fixed: [9967447](https://github.com/databricks/databricks-sdk-java/pull/324/commits/99674479030cbdad304b72217d1ddf9e89eda503)

databricks-sdk-java/src/main/java/com/databricks/sdk/mixin/JobsExt.java
ricardoc-db: > Ditto here

I may have missed another comment of yours.
I added a docstring to this method in this commit: [e9eec92](https://github.com/databricks/databricks-sdk-java/pull/324/commits/e9eec92a12a25950a5a71478f538e16ee38043b0) Is this what you were looking for?

ricardoc-db:
> Were you able to test this out against a real workspace?

Did not attempt.

databricks-sdk-java/src/main/java/com/databricks/sdk/mixin/JobsExt.java
ricardoc-db: Ah, you were referring to this? https://github.com/databricks/databricks-sdk-py/pull/725#discussion_r1718291278

mgyucht (approved):

ricardoc-db:
@mgyucht tested manually against a job with 100+ tasks as well as a ForEach run with 100+ iterations, hardcoding both v 2.1 and 2.2. All combinations worked as expected.

vsamoilov (approved):

renaudhartert-db:
Removing myself from the reviewer list as it seems my review is not explicitly needed/requested. Please add me back if that is incorrect.

PR #325, opened 2024-07-26
[Internal] Add integration tests for Jobs API

ricardoc-db:
## Changes

Add integration tests for Jobs API

## Tests

The tests are the change.


dvkuznietsov (approved):

renaudhartert-db (approved):
Approved.

Context: As we discussed, we're moving to a new framework to handle integration tests. However, given that (i) this framework is not ready, (ii) that you've taken the time to write these tests, and (iii) that they are very minimalist, I think it is better to proceed with this PR and adapt the test later on.

PR #326, opened 2024-07-30
[Internal] Fixed SqlIT integration tests

tanmay-db:
## Changes

Queries return null, we need to add a check to iterate if non null. 

Note: Ideally the test should create the query history and then fetch those followed by verifying if it's the same data to properly check for E2E lifecycle.

But for just listing, the requests work (although get null response) and more tests should be added for checking create etc...

## Tests

Ran Integration test locally.

### DIFF
--- a/databricks-sdk-java/src/test/java/com/databricks/sdk/integration/SqlIT.java
+++ b/databricks-sdk-java/src/test/java/com/databricks/sdk/integration/SqlIT.java
@@ -3,8 +3,10 @@ package com.databricks.sdk.integration;
 import com.databricks.sdk.WorkspaceClient;
 import com.databricks.sdk.integration.framework.EnvContext;
 import com.databricks.sdk.integration.framework.EnvTest;
-import com.databricks.sdk.service.sql.*;
-
+import com.databricks.sdk.service.sql.ListQueryHistoryRequest;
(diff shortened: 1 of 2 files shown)


renaudhartert-db (approved):
The change looks good to me but I'm wondering: was that test passing before? And if so, what changed to return null queries? Just want to be sure that this is not the sign of an actual regression.

tanmay-db:
We would need to put a note in the release notes about moving to new APIs but I think it should be okay to go ahead with merging this and releasing.

databricks-sdk-java/src/test/java/com/databricks/sdk/integration/SqlIT.java
```diff
@@ -22,8 +22,10 @@ void listQueryHistoryTimeRange(WorkspaceClient w) {
         new ListQueryHistoryRequest()
             .setFilterBy(new QueryFilter().setQueryStartTimeRange(timeRange));
     Iterable<QueryInfo> queries = w.queryHistory().list(request).getRes();
-    for (QueryInfo query : queries) {
-      System.out.println(query);
+    if (queries != null) {
```
tanmay-db, line 25: Note: We can't use the legacy query API here since `ListQueriesLegacyRequest` doesn't support `setFilterBy`

PR #331, opened 2024-08-12
[Feature] Add a way to provide non proxy hosts

vikrantpuppala:
## Changes

Add a new config in proxy hosts to provide non proxy hosts. Related system property info here: https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html

## Tests

Tested locally and via unit tests


databricks-sdk-java/src/main/java/com/databricks/sdk/core/ProxyConfig.java
```diff
@@ -7,6 +7,8 @@ public class ProxyConfig {
   private String password;
   private ProxyAuthType proxyAuthType;
   private Boolean useSystemProperties;
+  // Needs the nonProxyHosts to be set in the format of "suffix1,suffix2,suffix3"
```
mgyucht, line 10: Let's use | to separate hosts rather than , to follow the existing system property standard (https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html)

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ProxyConfig.java
```diff
@@ -80,4 +82,13 @@ public ProxyConfig setUseSystemProperties(Boolean useSystemProperties) {
     this.useSystemProperties = useSystemProperties;
     return this;
   }
+
+  public String getNonProxyHosts() {
+    return nonProxyHosts;
+  }
+
+  public ProxyConfig setNonProxyHosts(String nonProxyHosts) {
```
mgyucht, line 90: Add a doccomment for this.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/ProxyUtils.java
```diff
@@ -142,4 +158,29 @@ public static void setupBasicProxyAuth(
         .setDefaultCredentialsProvider(credsProvider)
         .setProxyAuthenticationStrategy(new ProxyAuthenticationStrategy());
   }
+
+  static class CustomRoutePlanner implements HttpRoutePlanner {
```
mgyucht, line 162: Can we move this to its own file if it doesn't have any dependencies in the parent class?

mgyucht, line 162: Let's add a doccomment here too.

mgyucht (changes requested):
Couple changes but generally think the approach is good.

mgyucht (approved):

PR #333, opened 2024-08-20
[Feature] Provide SSLSocketFactory in HttpClient

vikrantpuppala:
## Changes

We need to provide a way for clients to specify their own SSL socket factory in case they want to handle TLS connections in a different way

## Tests

Unit tests and local testing


mgyucht (changes requested):
Can we use builder pattern for constructing CommonsHttpClient?

mgyucht (commented):
Can you add doccomments? Otherwise LGTM.

mgyucht (approved):

PR #340, opened 2024-09-03
[Fix] Handle login.html incorrect validation for private link

satviksr-db:
## Changes
1. query returned can be null, however its not handled correctly
2. checking if query is null or not before checking for error code `private-link-validation-error`

## Tests
Unit Tests


avijeet95 (approved):

helfer (commented):
I don't know the codebase, but this would fix the issue we're seeing in our db-sentinel probes for the login page at https://databricks.atlassian.net/browse/ES-1247829. Thank you @satviksr-db!

mgyucht (approved):

PR #355, opened 2024-10-02
[Fix] Retry on too many auth requests

rauchy:
## Changes

This PR addresses an issue encountered during authentication, where the server returns a 429 status code when fetching OIDC endpoints from a well-known location, particularly during frequent authentication requests. The update introduces a retry mechanism within the auth flow to mitigate the impact of 429 errors.

Modified `ApiClient` object construction to allow unauthenticated calls by including authentication function injection, where within the OIDC workflow act as no-op actions. `DatabricksConfig` would now instantiate an `ApiClient` instead of a standard `CommonsHttpClient` in these cases.
errors during a complete authentication workflow, especially under high concurrent request conditions.

## Tests

* Added unit tests to verify the retry functionality upon encountering a 429 response.
* Included a manual test to simulate real-world scenarios, ensuring the retry mechanism effectively handles 429

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
@@ -38,32 +38,32 @@ public class ApiClient {
     private Function<Void, String> getAuthTypeFunc;
 
     public Builder withDatabricksConfig(DatabricksConfig config) {
-        this.config = config;
-        return this;
+      this.config = config;
+      return this;
     }
 
     public Builder withTimer(Timer timer) {
-        this.timer = timer;
-        return this;
+      this.timer = timer;
+      return this;
     }
 
     public Builder withAuthenticateFunc(Function<Void, Map<String, String>> authenticateFunc) {
-        this.authenticateFunc = authenticateFunc;
-        return this;
+      this.authenticateFunc = authenticateFunc;
+      return this;
     }
 
     public Builder withGetHostFunc(Function<Void, String> getHostFunc) {
-        this.getHostFunc = getHostFunc;
-        return this;
+      this.getHostFunc = getHostFunc;
+      return this;
     }
 
     public Builder withGetAuthTypeFunc(Function<Void, String> getAuthTypeFunc) {
-        this.getAuthTypeFunc = getAuthTypeFunc;
-        return this;
+      this.getAuthTypeFunc = getAuthTypeFunc;
+      return this;
     }
 
     public ApiClient build() {
-        return new ApiClient(config, timer, authenticateFunc, getHostFunc, getAuthTypeFunc);
+      return new ApiClient(config, timer, authenticateFunc, getHostFunc, getAuthTypeFunc);
     }
   }
 
@@ -99,10 +99,20 @@ public class ApiClient {
   }
 
   public ApiClient(DatabricksConfig config, Timer timer) {
-    this(config, timer, v -> config.authenticate(), v -> config.getHost(), v -> config.getAuthType());
+    this(
+        config,
+        timer,
+        v -> config.authenticate(),
+        v -> config.getHost(),
+        v -> config.getAuthType());
   }
 
-  public ApiClient(DatabricksConfig config, Timer timer, Function<Void, Map<String, String>> authenticateFunc, Function<Void, String> getHostFunc, Function<Void, String> getAuthTypeFunc) {
+  public ApiClient(
+      DatabricksConfig config,
+      Timer timer,
+      Function<Void, Map<String, String>> authenticateFunc,
+      Function<Void, String> getHostFunc,
+      Function<Void, String> getAuthTypeFunc) {
     this.config = config;
     this.timer = timer;
     this.authenticateFunc = authenticateFunc;
@@ -127,7 +137,7 @@ public class ApiClient {
     httpClient = config.getHttpClient();
     bodyLogger = new BodyLogger(mapper, 1024, debugTruncateBytes);
     retryStrategyPicker = new RequestBasedRetryStrategyPicker(this.config);
-    }
+  }
 
   private static <I> void setQuery(Request in, I entity) {
     if (entity == null) {
@@ -305,7 +315,7 @@ public class ApiClient {
       String authType = getAuthTypeFunc.apply(null);
       String userAgent = UserAgent.asString();
       if (authType != "") {
-          userAgent += String.format(" auth/%s", authType);
+        userAgent += String.format(" auth/%s", authType);
       }
       in.withHeader("User-Agent", userAgent);
 
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
@@ -320,9 +320,9 @@ public class ApiClient {
 
       // Set User-Agent with auth type info, which is available only
       // after the first invocation to config.authenticate()
-      String authType = getAuthTypeFunc.apply(null);
       String userAgent = UserAgent.asString();
-      if (authType != "") {
+      if (getAuthTypeFunc != null) {
+        String authType = getAuthTypeFunc.apply(null);
         userAgent += String.format(" auth/%s", authType);
       }
       in.withHeader("User-Agent", userAgent);
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
@@ -31,14 +31,26 @@ import org.slf4j.LoggerFactory;
  */
 public class ApiClient {
   static class Builder {
-    private DatabricksConfig config;
     private Timer timer;
     private Function<Void, Map<String, String>> authenticateFunc;
     private Function<Void, String> getHostFunc;
     private Function<Void, String> getAuthTypeFunc;
+    private Integer debugTruncateBytes;
+    private HttpClient httpClient;
+    private String accountId;
+    private RetryStrategyPicker retryStrategyPicker;
+    private boolean isDebugHeaders;
 
     public Builder withDatabricksConfig(DatabricksConfig config) {
-      this.config = config;
+      this.authenticateFunc = v -> config.authenticate();
+      this.getHostFunc = v -> config.getHost();
+      this.getAuthTypeFunc = v -> config.getAuthType();
+      this.debugTruncateBytes = config.getDebugTruncateBytes();
+      this.httpClient = config.getHttpClient();
+      this.accountId = config.getAccountId();
+      this.retryStrategyPicker = new RequestBasedRetryStrategyPicker(config.getHost());
+      this.isDebugHeaders = config.isDebugHeaders();
+
       return this;
     }
 
@@ -62,8 +74,18 @@ public class ApiClient {
       return this;
     }
 
+    public Builder withHttpClient(HttpClient httpClient) {
+      this.httpClient = httpClient;
+      return this;
+    }
+
+    public Builder withRetryStrategyPicker(RetryStrategyPicker retryStrategyPicker) {
+      this.retryStrategyPicker = retryStrategyPicker;
+      return this;
+    }
+
     public ApiClient build() {
-      return new ApiClient(config, timer, authenticateFunc, getHostFunc, getAuthTypeFunc);
+      return new ApiClient(this);
     }
   }
 
@@ -72,8 +94,6 @@ public class ApiClient {
   private final int maxAttempts;
 
   private final ObjectMapper mapper;
-
-  private final DatabricksConfig config;
 
   private final Random random;
 
@@ -84,6 +104,8 @@ public class ApiClient {
   private final Function<Void, Map<String, String>> authenticateFunc;
   private final Function<Void, String> getHostFunc;
   private final Function<Void, String> getAuthTypeFunc;
+  private final String accountId;
+  private final boolean isDebugHeaders;
   private static final String RETRY_AFTER_HEADER = "retry-after";
 
   public ApiClient() {
@@ -91,7 +113,7 @@ public class ApiClient {
   }
 
   public String configuredAccountID() {
-    return config.getAccountId();
+    return accountId;
   }
 
   public ApiClient(DatabricksConfig config) {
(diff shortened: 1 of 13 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -58,7 +99,26 @@ public ApiClient(DatabricksConfig config) {
...
+        v -> config.getHost(),
+        v -> config.getAuthType());
+  }
+
+  public ApiClient(
+      DatabricksConfig config,
+      Timer timer,
+      Function<Void, Map<String, String>> authenticateFunc,
+      Function<Void, String> getHostFunc,
+      Function<Void, String> getAuthTypeFunc) {
```
mgyucht, line 115: private ApiClient(Builder builder) {...}

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -58,7 +99,26 @@ public ApiClient(DatabricksConfig config) {
   }
 
   public ApiClient(DatabricksConfig config, Timer timer) {
+    this(
+        config,
+        timer,
+        v -> config.authenticate(),
+        v -> config.getHost(),
+        v -> config.getAuthType());
```
mgyucht, line 107: this(new Builder().withConfig(config.resolve()).withTimer(timer))

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -29,6 +30,43 @@
...
+      return this;
+    }
+
+    public Builder withGetAuthTypeFunc(Function<Void, String> getAuthTypeFunc) {
+      this.getAuthTypeFunc = getAuthTypeFunc;
+      return this;
+    }
+
+    public ApiClient build() {
+      return new ApiClient(config, timer, authenticateFunc, getHostFunc, getAuthTypeFunc);
```
mgyucht, line 66: return new ApiClient(this);

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -58,7 +99,26 @@ public ApiClient(DatabricksConfig config) {
   }
 
   public ApiClient(DatabricksConfig config, Timer timer) {
+    this(
+        config,
+        timer,
+        v -> config.authenticate(),
```
mgyucht, line 105: config::authenticate

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -29,6 +30,43 @@
  * guessing
  */
 public class ApiClient {
+  static class Builder {
+    private DatabricksConfig config;
+    private Timer timer;
+    private Function<Void, Map<String, String>> authenticateFunc;
```
mgyucht, line 36: Can we use Supplier for these?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -245,15 +304,19 @@ private Response executeInner(Request in, String path) {
...
       // This call may configure the host (e.g. in case of notebook native auth).
-      in.withUrl(config.getHost() + path);
+      in.withUrl(getHostFunc.apply(null) + path);
 
       // Set User-Agent with auth type info, which is available only
       // after the first invocation to config.authenticate()
-      String userAgent = String.format("%s auth/%s", UserAgent.asString(), config.getAuthType());
+      String authType = getAuthTypeFunc.apply(null);
+      String userAgent = UserAgent.asString();
+      if (authType != "") {
```
mgyucht, line 317: If getAuthTypeFunc is null, we can jst check if that func is null

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
rauchy: Suppliers look good for replacing the use of `Function` here, but some functions (`authenticate` as an example) are overloaded, and Supplier and the `::` syntax can't deal with those. So if we choose to use Supplier here, _some_ functions will still be stored under a `Function` variable, which kinda makes it even more complicated :/

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
rauchy: 59f49977

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
rauchy: See https://github.com/databricks/databricks-sdk-java/pull/355#discussion_r1786060249

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
rauchy: https://github.com/databricks/databricks-sdk-java/pull/355/commits/652cb0b7c629e2b4a38725ea4069a874352eec88

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
rauchy: https://github.com/databricks/databricks-sdk-java/pull/355/commits/652cb0b7c629e2b4a38725ea4069a874352eec88

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
rauchy: https://github.com/databricks/databricks-sdk-java/pull/355/commits/652cb0b7c629e2b4a38725ea4069a874352eec88

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -29,55 +30,119 @@
...
+    private Function<Void, Map<String, String>> authenticateFunc;
+    private Function<Void, String> getHostFunc;
+    private Function<Void, String> getAuthTypeFunc;
+    private Integer debugTruncateBytes;
+    private HttpClient httpClient;
+    private String accountId;
+    private RetryStrategyPicker retryStrategyPicker;
+    private boolean isDebugHeaders;
+
+    public Builder withDatabricksConfig(DatabricksConfig config) {
```
mgyucht, line 44: Actually, this could even be private, since it is only used in the default constructor.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -202,6 +267,8 @@ private <I> Request prepareBaseRequest(String method, String path, I in)
     } else if (InputStream.class.isAssignableFrom(in.getClass())) {
       InputStream body = (InputStream) in;
       return new Request(method, path, body);
+    } else if (in instanceof String) {
```
mgyucht, line 270: Suggestion: leave a comment that this assumes that the request is already serialized.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -29,55 +30,119 @@
...
-    config.resolve();
+    this(new Builder().withDatabricksConfig(config.resolve()).withTimer(timer));
+  }
 
-    Integer rateLimit = config.getRateLimit();
-    if (rateLimit == null) {
-      rateLimit = 15;
-    }
+  private ApiClient(Builder builder) {
+    this.timer = builder.timer;
```
mgyucht, line 128: Suggestion: let's use sane defaults when those values are not provided in the builder. Ideally `ApiClient.Builder().build()` would make a more or less working ApiClient. Here, we can use `SystemTimer()` if not specified.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -581,12 +582,19 @@ private OpenIDConnectEndpoints fetchDefaultOidcEndpoints() throws IOException {
...
 
-    String oidcEndpoint = getHost() + "/oidc/.well-known/oauth-authorization-server";
-    Response resp = getHttpClient().execute(new Request("GET", oidcEndpoint));
-    if (resp.getStatusCode() != 200) {
-      return null;
-    }
-    return new ObjectMapper().readValue(resp.getBody(), OpenIDConnectEndpoints.class);
+    ApiClient apiClient =
+        new ApiClient.Builder()
+            .withDatabricksConfig(this)
```
mgyucht, line 587: Do we need to specify the config directly here?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -581,12 +582,19 @@ private OpenIDConnectEndpoints fetchDefaultOidcEndpoints() throws IOException {
...
+            .withTimer(new SystemTimer())
+            .withAuthenticateFunc(v -> new HashMap<String, String>())
+            .withGetHostFunc(v -> getHost())
+            .withGetAuthTypeFunc(v -> "")
```
mgyucht, line 591: This we should probably leave out? Otherwise it is not null according to your check.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -245,15 +312,19 @@ private Response executeInner(Request in, String path) {
       Response out = null;
 
       // Authenticate the request. Failures should not be retried.
-      in.withHeaders(config.authenticate());
+      in.withHeaders(authenticateFunc.apply(null));
```
mgyucht, line 315: Suggestion: we can treat `authenticateFunc` as nullable to indicate whether headers need to be added. Thus:
```java
if (authenticateFunc != null) {
    in.withHeaders(authenticateFunc.apply(null));
}
```
This way, users don't have to supply an empty authenticateFunc in the builder.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -581,12 +582,19 @@ private OpenIDConnectEndpoints fetchDefaultOidcEndpoints() throws IOException {
...
-    Response resp = getHttpClient().execute(new Request("GET", oidcEndpoint));
-    if (resp.getStatusCode() != 200) {
-      return null;
-    }
-    return new ObjectMapper().readValue(resp.getBody(), OpenIDConnectEndpoints.class);
+    ApiClient apiClient =
+        new ApiClient.Builder()
+            .withDatabricksConfig(this)
+            .withTimer(new SystemTimer())
+            .withAuthenticateFunc(v -> new HashMap<String, String>())
```
mgyucht, line 589: If you follow my suggestion from above, we can leave this out.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -62,17 +62,27 @@ protected static Token retrieveToken(
...
     try {
-      Response rawResp = hc.execute(req);
-      OAuthResponse resp = new ObjectMapper().readValue(rawResp.getBody(), OAuthResponse.class);
+      ApiClient apiClient =
+          new ApiClient.Builder()
+              .withHttpClient(hc)
+              .withTimer(new SystemTimer())
+              .withAuthenticateFunc(v -> headers)
+              .withGetAuthTypeFunc(v -> "")
+              .withGetHostFunc(v -> "")
```
mgyucht, line 73: ditto, these can be removed if you take my suggestions

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -29,55 +30,119 @@
...
-      rateLimit = 15;
-    }
+  private ApiClient(Builder builder) {
+    this.timer = builder.timer;
+    this.authenticateFunc = builder.authenticateFunc;
+    this.getHostFunc = builder.getHostFunc;
+    this.getAuthTypeFunc = builder.getAuthTypeFunc;
+    this.httpClient = builder.httpClient;
+    this.accountId = builder.accountId;
+    this.retryStrategyPicker = builder.retryStrategyPicker;
```
mgyucht, line 134: Same here, can we use the default request-based retry strategy picker if none is provided?

databricks-sdk-java/src/test/java/com/databricks/sdk/DatabricksAuthLoadTest.java
```diff
@@ -0,0 +1,73 @@
+// These are auto-generated tests for Unified Authentication
```
mgyucht, line 1: This is not autogenerated?

databricks-sdk-java/src/test/java/com/databricks/sdk/DatabricksAuthLoadTest.java
```diff
@@ -0,0 +1,73 @@
...
+import com.databricks.sdk.core.ConfigResolving;
+import com.databricks.sdk.core.DatabricksConfig;
+import com.databricks.sdk.core.commons.CommonsHttpClient;
+import com.databricks.sdk.core.utils.GitHubUtils;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.*;
+import org.junit.jupiter.api.Test;
+
+public class DatabricksAuthLoadTest implements GitHubUtils, ConfigResolving {
```
mgyucht, line 17: Let's leave a comment here about this test.

Also small suggestion: let's move this into a `benchmark` package. It'll be a new convention within Java SDK tests to keep things a bit more organized for now.

mgyucht (commented):
Looking good. Few comments, otherwise we're on the right track.

databricks-sdk-java/src/test/java/com/databricks/sdk/DatabricksAuthLoadTest.java
rauchy: https://github.com/databricks/databricks-sdk-java/pull/355/commits/68ae507c8f0888d60d5bcac44c809413aa5dec18

databricks-sdk-java/src/test/java/com/databricks/sdk/DatabricksAuthLoadTest.java
rauchy: https://github.com/databricks/databricks-sdk-java/pull/355/commits/68ae507c8f0888d60d5bcac44c809413aa5dec18

rauchy:
Great comments @mgyucht, thanks for the thorough review.

The null checks make sense. I eventually went with a more "tell, don't ask" approach by providing no-ops in the private `ApiClient` constructor. This way we could avoid null paranoia.

mgyucht (approved):
LGTM. Make sure to run `make fmt` to fix the format check.

I will take a look at how Maven is running tests to see why this test is being executed. In the meantime, we should be able to use the Disabled annotation (https://junit.org/junit5/docs/5.2.0/api/org/junit/jupiter/api/Disabled.html) to disable the test explicitly.

PR #357, opened 2024-10-04
[Fix] Make `UserAgent`'s `otherInfo` thread-safe

mgyucht:
## Changes
`UserAgent.asString()` is called by `ApiClient` every time a request is made. `UserAgent` contains an `otherInfo` list of additional metadata that users can add to the user agent as needed. If users concurrently modify this `otherInfo` value while iterating through it when making a request, `ConcurrentModificationException` is thrown.

To prevent this, we add `synchronized` blocks around modifying and reading from `otherInfo`. Since the time spent constructing the user agent is generally fast, this should not significantly affect performance.

I also added a readme describing the new benchmark module with a small contribution guide, and I renamed the existing load test to align with that contribution guide.

## Tests
Added a `UserAgentLoadTest` to validate that you can safely add to `otherInfo` and make requests.

Added a `WorkspaceClientLoadIT` to exercise the SDK end-to-end with the SCIM Me request, concurrently.


databricks-sdk-java/src/test/java/com/databricks/sdk/benchmark/DatabricksAuthLoadIT.java
```diff
@@ -2,20 +2,35 @@
...
+public class DatabricksAuthLoadIT {
 
   @Test
-  @Disabled
-  public void testConcurrentConfigBasicAuthAttrs() throws Exception {
+  public void testConcurrentConfigBasicAuthAttrs(
+      @EnvOrSkip("DATABRICKS_HOST") String host,
+      @EnvOrSkip("DATABRICKS_CLIENT_ID") String clientId,
+      @EnvOrSkip("DATABRICKS_CLIENT_SECRET") String clientSecret)
+      throws Exception {
```
rauchy, line 33: Much better! :)

rauchy (approved):

PR #360, opened 2024-10-11
[Fix] Make userAgent compatible with SQL Drivers

samikshya-db:
## Changes
- Currently user Agent does not work with SQL Driver flows. This PR fixes it

## Tests
- Verifed that LogRedactor populates values correctly 
- Added unit tests in this repo too
- Verified end to end flow using local SDK Jar


renaudhartert-db:
Closing this PR as per offline discussions: use agent should be sanitized.

renaudhartert-db:
Closing this PR as per offline discussions: use agent should be sanitized.

PR #363, opened 2024-10-17
[Feature] Experimental: Make retry strategy configurable

avijeet95:
Adds support to configure Retry Strategy in HttpClient. Currently only the default retry strategy is used. This strategy retries 3 times and does not have any sleep interval in between. For our use case, we would prefer using the ExponentialBackOffStrategy.

This does not affect the default behavior but gives additional options to the users.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/commons/CommonsHttpClient.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.stream.Collectors;
 import org.apache.commons.io.IOUtils;
 import org.apache.http.*;
+import org.apache.http.client.HttpRequestRetryHandler;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.*;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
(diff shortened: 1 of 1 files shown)


satviksr-db (approved):

mgyucht (approved):
Let's label this as experimental, but otherwise this is fine for me.

Long-term, I think we will want to implement retries at a higher layer, after the response is deserialized, so that the retry implementation can review a deserialized error response before making a decision.

avijeet95:
> Let's label this as experimental, but otherwise this is fine for me.
> 
> Long-term, I think we will want to implement retries at a higher layer, after the response is deserialized, so that the retry implementation can review a deserialized error response before making a decision.

Yes right now there are retries from the SDK level and then further in HttpClient. But making that decision is a bit more involved so this PR unblocks us till we get to the point you suggested. Thanks!

avijeet95:
Added experimental to list of valid tags

mgyucht:
Ah sorry, I meant to label `withRequestRetryHandler` as experimental in a doccomment. The PR tags need to be consistent across all SDKs to minimize unnecessary variance. Let's set the PR title to`[Feature] Experimental: Make retry strategy configurable`.

murtadhask:
> Let's label this as experimental, but otherwise this is fine for me.
> 
> Long-term, I think we will want to implement retries at a higher layer, after the response is deserialized, so that the retry implementation can review a deserialized error response before making a decision.

isn't that already implemented in RetryStrategyPicker?
I'm confused now.

PR #374, opened 2024-10-31
[Internal] Always write message for manual test integration

hectorcast-db:
## Changes
Old script could not be run from master due to security restrictions and there is no reliable way to detect if a user as secrets.

## Tests
Run as pull_request.

### DIFF
--- a/.github/workflows/external-message.yml
+++ b/.github/workflows/external-message.yml
@@ -11,7 +11,6 @@ on:
     branches:
       - main
 
-
 jobs:
   comment-on-pr:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/external-message.yml
```diff
@@ -11,81 +11,53 @@ on:
...
+      - uses: actions/checkout@v4
+
+      # We cannot check if the user has access to secrets from this workflow.
+      # However, if they user has access to secrets, the integration-tests.yml workflow will 
+      # write a message. Wait and check if the message is present.
+      # The results can be wrong if the worker pool is out of capacity and the integration-tests.yml 
+      # workflow runs after this workflow. In that case, we will get a message
+      # request for manual test execution, which is misleading since
+      # the test will still run automatically.
+      - name: Wait for 60 seconds
```
parthban-db, line 30: Can we increase this time? 1 min doesn't seem to be enough.

parthban-db (approved):

renaudhartert-db:
Anything that prevents us from showing the comment to everyone and making it clear (in the comment) that it is intended for the reviewers?

.github/workflows/integration-tests.yml
```diff
@@ -10,17 +10,18 @@ jobs:
     check-token:
         name: Check secrets access
         runs-on: ubuntu-latest
+        environment: "test-trigger-is"
         outputs:
             has_token: ${{ steps.set-token-status.outputs.has_token }}
         steps:
-        - name: Check if GITHUB_TOKEN is set
+        - name: Check if DECO_WORKFLOW_TRIGGER_APP_ID is set
```
hectorcast-db, line 17: GITHUB_TOKEN is sometimes set for external contributors but with a smaller set of permissions. It is not reliable.

renaudhartert-db (approved):
LGTM

PR #376, opened 2024-11-02
[Feature] DatabricksConfig: Add clone() support

satviksr-db:
## Changes
Adds support for cloning DatabricksConfig(), this is needed because we need a way to set configurations per API call such as 
1. timeout
2. httpClient configuration
3. debugHeaders

However, we still want to use the cached oauth token etc and we would like the header factory to be a common object across workspace clients and databricks configs

Ideally this should be supported by the SDK itself natively, however since its not supported and it will take significant migration to achieve that, this is a work around to achieve the same.  

## Tests
Added UT


hectorcast-db (approved):

hectorcast-db:
Hi @satviksr-db 
Thanks for the contribution. Approving and merging. 
This will released in the next release.

PR #386, opened 2024-11-08
[Internal] Update Jobs GetRun API to support paginated responses for jobs and ForEach tasks

gkiko10:

Introduces extension for jobs getRun call that paginates tasks and iterations arrays in the response and returns aggregated response to the caller. This change is necessary to prepare for jobs API 2.2 release that serves paginated response. Pagination is over once the next_page_token is absent from the response. The pagination logic is not exposed to the customer.

## How is this tested?

Unit tests


renaudhartert-db (changes requested):
Could you undo the changes in `WorkspaceClient.java`? These will be done automatically once we have updated the code generation template.

renaudhartert-db (approved):

PR #401, opened 2025-01-22
[Fix] Correctly include query parameters for APIs whose request objects contain the body as a field

mgyucht:
The Databricks API spec has two different styles of APIs, those where the request body is a field in the `Request` structure, and those where the request body is said structure (not counting the parameters passed via the URI path and query string). The former style of APIs was incorrectly implemented, preventing such APIs from using any query parameters.

Generally, the approach taken here is to move the construction of the request object into the autogenerated `Impl` classes and providing a fully materialized `Request` object to the `ApiClient`. To do so, we expose the `<T> T execute(Request in, Target Class<T>)` method as public.

Separately, I've removed the per-HTTP-method methods from `ApiClient`, opting instead to use the singular `execute`, as they purely create more maintenance burden for us and decrease the surface area we need to maintain.

## How is this tested?
Added `AppsImplTest` to verify that the no_compute flag is included in the request passed to the `ApiClient`.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
@@ -198,65 +198,10 @@ public class ApiClient {
     }
   }
 
-  public <O> O HEAD(String path, Class<O> target, Map<String, String> headers) {
-    return HEAD(path, null, target, headers);
-  }
-
-  public <I, O> O HEAD(String path, I in, Class<O> target, Map<String, String> headers) {
+  public <I, O> O execute(
+      String method, String path, I in, Class<O> target, Map<String, String> headers) {
     try {
-      return execute(prepareRequest("HEAD", path, in, headers), target);
-    } catch (IOException e) {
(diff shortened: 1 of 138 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -198,65 +198,10 @@ protected <I, O> O withJavaType(
...
-  public <I, O> O PATCH(String path, I in, Class<O> target, Map<String, String> headers) {
-    try {
-      return execute(prepareRequest("PATCH", path, in, headers), target);
-    } catch (IOException e) {
-      throw new DatabricksException("IO error: " + e.getMessage(), e);
-    }
-  }
-
-  public <I, O> O DELETE(String path, I in, Class<O> target, Map<String, String> headers) {
+  public <I, O> O execute(
```
hectorcast-db, line 201: Do we still use this one? Now all generated code uses the one which takes the Request object, correct?

hectorcast-db (approved):

PR #423, opened 2025-03-25
Support Databricks Workload Identity Federation for GitHub tokens

hectorcast-db:

This PR adds support for Databricks Workload Identity Federation using GitHub tokens. This allows users to use WIF from their GitHub Workflows and authenticate their workloads without long lived secrets. 

This new credentials strategy is added to the DefaultCredentialsStrategy after the other Databricks Credentials Strategy and before cloud specific authentication methods.
WIF credentials uses a subset of configuration values of other Databricks authentication methods. By being added after them it ensures that WIF is not used when  other Databricks authentication methods are configured.
WIF uses the Databricks client id, which is not used by cloud specific authentication methods. Therefore, it will not be used when cloud specific authentication methods are configured.

## How is this tested?
Added tests.


databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/ClientCredentials.java
```diff
@@ -59,34 +61,33 @@ public Builder withHttpClient(HttpClient hc) {
 
     public ClientCredentials build() {
       Objects.requireNonNull(this.clientId, "clientId must be specified");
-      Objects.requireNonNull(this.clientSecret, "clientSecret must be specified");
```
hectorcast-db, line 62: Not required for WIF/OIDC

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksWifCredentialsProvider.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * DatabricksWIFCredentials uses a Token Supplier to get a JWT Token and exchanges it for a
+ * Databricks Token. Supported suppliers: - GitHub OIDC
+ */
+public class DatabricksWifCredentialsProvider implements CredentialsProvider {
```
renaudhartert-db, line 17: Should this be `GithubOidcCredentialsProvider`?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/GitHubOidcTokenSupplier.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.IOException;
+
+public class GitHubOidcTokenSupplier {
+
+  private final ObjectMapper mapper = new ObjectMapper();
+
+  private final DatabricksConfig config;
+
+  public GitHubOidcTokenSupplier(DatabricksConfig config) {
+    this.config = config;
```
renaudhartert-db, line 18: Could we rather set the specific variables that we need from the config? This will help decouple the internal logic from the config.

renaudhartert-db (changes requested):

renaudhartert-db (approved):

PR #433, opened 2025-04-23
Add IDTokenSource Interface

emmyzhou-db:
This PR adds the foundational components for OIDC support in the Databricks Java SDK:

**IDToken Class:** Encapsulates ID tokens from identity providers for OAuth flows.
**IDTokenSource** Interface: Defines a standard for retrieving ID tokens.

## How is this tested?
- Unit tests added for the IDToken class.
- IDTokenSource is not tested, as no implementation exists yet.

NO_CHANGELOG=true

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/IDToken.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/IDToken.java
@@ -1,22 +1,23 @@
 package com.databricks.sdk.core.oauth;
 
 /**
- * Represents an ID Token provided by an identity provider from an OAuth flow.
- * This token can later be exchanged for an access token.
+ * Represents an ID Token provided by an identity provider from an OAuth flow. This token can later
+ * be exchanged for an access token.
  */
 public class IDToken {
-    // The string value of the ID Token
-    private final String value;
+  // The string value of the ID Token
+  private final String value;
 
-    /**
-     * Constructs an IDToken with a value.
-     * @param value The ID Token string.
-     */
-    public IDToken(String value) {
-        this.value = value;
-    }
+  /**
+   * Constructs an IDToken with a value.
+   *
+   * @param value The ID Token string.
+   */
+  public IDToken(String value) {
+    this.value = value;
+  }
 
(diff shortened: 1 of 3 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/IDToken.java
```diff
@@ -0,0 +1,23 @@
...
+  /**
+   * Constructs an IDToken with a value.
+   *
+   * @param value The ID Token string.
+   */
+  public IDToken(String value) {
+    this.value = value;
+  }
+
+  public String getValue() {
```
renaudhartert-db, line 20: Because I feel like I have to say something, could you please add a small comment on this method?

renaudhartert-db (approved):
LGTM modulo open comment

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/IDToken.java
```diff
@@ -0,0 +1,23 @@
...
+  private final String value;
+
+  /**
+   * Constructs an IDToken with a value.
+   *
+   * @param value The ID Token string.
+   */
+  public IDToken(String value) {
+    this.value = value;
+  }
```
parthban-db, line 18: Add a null check in the constructor.
```suggestion
  public IDToken(String value) {
    if (value == null || value.isEmpty()) {
        throw new IllegalArgumentException("ID Token value cannot be null or empty");
    }
    this.value = value;
  }
```

databricks-sdk-java/src/test/java/com/databricks/sdk/core/oauth/IDTokenTest.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class IDTokenTest {
+
+  private static final String accessToken = "testIdToken";
+
+  @Test
+  void createIDToken() {
+    IDToken idToken = new IDToken(accessToken);
+    assertEquals(accessToken, idToken.getValue());
+  }
```
parthban-db, line 15: Add test for the null value.

PR #439, opened 2025-04-30
Add DatabricksOAuthTokenSource

emmyzhou-db:

- Implemented DatabricksOAuthTokenSource class that handles the OAuth token exchange flow
- This class manages the OAuth token exchange flow using ID tokens to obtain access tokens
- The token exchange mechanism is essential for implementing OIDC-based authentication in Databricks

## How is this tested?
The implementation is tested through unit tests that:

- Mock the ID Token Source to simulate the token exchange flow
- Mock the HTTP client to verify token exchange requests and responses
- Test the complete token exchange flow including audience selection and error handling
- Validate proper parameter handling and error responses
- All tests are automated and part of the unit test suite. No manual testing was required.

NO_CHANGELOG=true

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
@@ -1,0 +1,200 @@
+package com.databricks.sdk.core.oauth;
+
+import com.databricks.sdk.core.DatabricksException;
+import com.databricks.sdk.core.http.FormRequest;
+import com.databricks.sdk.core.http.HttpClient;
+import com.databricks.sdk.core.http.Response;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Strings;
+import java.io.IOException;
+import java.time.LocalDateTime;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Implementation of TokenSource that handles OAuth token exchange for Databricks authentication.
+ * This class manages the OAuth token exchange flow using ID tokens to obtain access tokens.
+ */
+public class DatabricksOAuthTokenSource implements TokenSource {
+  /** OAuth client ID used for token exchange */
+  private final String clientId;
+  /** Databricks account ID, used as audience if provided */
+  private final String accountId;
+  /** OpenID Connect endpoints configuration */
+  private final OpenIDConnectEndpoints endpoints;
+  /** Custom audience value for token exchange */
+  private final String audience;
+  /** Source of ID tokens used in token exchange */
+  private final IDTokenSource idTokenSource;
+  /** HTTP client for making token exchange requests */
+  private final HttpClient httpClient;
+
+  private DatabricksOAuthTokenSource(Builder builder) {
+    this.clientId = builder.clientId;
+    this.accountId = builder.accountId;
+    this.endpoints = builder.endpoints;
+    this.audience = builder.audience;
+    this.idTokenSource = builder.idTokenSource;
+    this.httpClient = builder.httpClient;
+  }
+
+  /**
+   * Builder class for constructing DatabricksOAuthTokenSource instances. Provides a fluent
+   * interface for setting required and optional parameters.
+   */
+  public static class Builder {
+    private final String clientId;
+    private final String host;
+    private final OpenIDConnectEndpoints endpoints;
+    private final IDTokenSource idTokenSource;
+    private final HttpClient httpClient;
+    private String accountId;
+    private String audience;
+
+    /**
+     * Validates that a value is non-empty and non-null for required fields.
+     *
+     * @param value The value to validate
+     * @param fieldName The name of the field being validated
+     * @throws IllegalArgumentException if validation fails
+     */
+    private static void validate(Object value, String fieldName) {
+      if (value == null) {
+        throw new IllegalArgumentException(fieldName + " must be non-null");
+      }
+      if (value instanceof String && ((String) value).isEmpty()) {
+        throw new IllegalArgumentException(fieldName + " must be non-empty");
+      }
+    }
+
+    /**
+     * Creates a new Builder with required parameters.
+     *
+     * @param clientId OAuth client ID
+     * @param host Databricks host URL
+     * @param endpoints OpenID Connect endpoints configuration
+     * @param idTokenSource Source of ID tokens
+     * @param httpClient HTTP client for making requests
+     */
+    public Builder(
+        String clientId,
+        String host,
+        OpenIDConnectEndpoints endpoints,
+        IDTokenSource idTokenSource,
+        HttpClient httpClient) {
+      validate(clientId, "ClientID");
+      validate(host, "Host");
+      validate(endpoints, "Endpoints");
+      validate(idTokenSource, "IDTokenSource");
+      validate(httpClient, "HttpClient");
+
+      this.clientId = clientId;
+      this.host = host;
+      this.endpoints = endpoints;
+      this.idTokenSource = idTokenSource;
+      this.httpClient = httpClient;
+    }
+
+    /**
+     * Sets the Databricks account ID.
+     *
+     * @param accountId The account ID
+     * @return This builder instance
+     */
+    public Builder accountId(String accountId) {
+      validate(accountId, "AccountID");
+      this.accountId = accountId;
+      return this;
+    }
+
+    /**
+     * Sets a custom audience value for token exchange.
+     *
+     * @param audience The audience value
+     * @return This builder instance
+     */
+    public Builder audience(String audience) {
+      validate(audience, "Audience");
+      this.audience = audience;
+      return this;
+    }
+
+    /**
+     * Builds a new DatabricksOAuthTokenSource instance.
+     *
+     * @return A new DatabricksOAuthTokenSource
+     */
+    public DatabricksOAuthTokenSource build() {
+      return new DatabricksOAuthTokenSource(this);
+    }
+  }
+
+  /**
+   * Retrieves an OAuth token by exchanging an ID token. Implements the OAuth token exchange flow to
+   * obtain an access token.
+   *
+   * @return A Token containing the access token and related information
+   * @throws DatabricksException if token exchange fails
+   */
+  @Override
+  public Token getToken() {
+    String effectiveAudience = determineAudience();
+    IDToken idToken = idTokenSource.getIDToken(effectiveAudience);
+
+    Map<String, String> params = new HashMap<>();
+    params.put("grant_type", "urn:ietf:params:oauth:grant-type:token-exchange");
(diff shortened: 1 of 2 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,200 @@
...
+   */
+  @Override
+  public Token getToken() {
+    String effectiveAudience = determineAudience();
+    IDToken idToken = idTokenSource.getIDToken(effectiveAudience);
+
+    Map<String, String> params = new HashMap<>();
+    params.put("grant_type", "urn:ietf:params:oauth:grant-type:token-exchange");
+    params.put("subject_token", idToken.getValue());
+    params.put("subject_token_type", "urn:ietf:params:oauth:token-type:jwt");
```
parthban-db, line 147: Let's define these constants as static final variables in the class.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /**
+     * Creates a new Builder with required parameters.
+     *
+     * @param clientId OAuth client ID
+     * @param host Databricks host URL
+     * @param endpoints OpenID Connect endpoints configuration
+     * @param idTokenSource Source of ID tokens
+     * @param httpClient HTTP client for making requests
```
parthban-db, line 77: Let's document the error that this function can throw.
```suggestion
     * @param httpClient HTTP client for making requests
     * @throws IllegalArgumentException if any required parameter is null or empty
```

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,200 @@
...
+      this.idTokenSource = idTokenSource;
+      this.httpClient = httpClient;
+    }
+
+    /**
+     * Sets the Databricks account ID.
+     *
+     * @param accountId The account ID
+     * @return This builder instance
+     */
```
parthban-db, line 103: ditto

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,200 @@
...
+      return this;
+    }
+
+    /**
+     * Sets a custom audience value for token exchange.
+     *
+     * @param audience The audience value
+     * @return This builder instance
+     */
+    public Builder audience(String audience) {
```
parthban-db, line 116: ditto

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,200 @@
...
+      return new DatabricksOAuthTokenSource(this);
+    }
+  }
+
+  /**
+   * Retrieves an OAuth token by exchanging an ID token. Implements the OAuth token exchange flow to
+   * obtain an access token.
+   *
+   * @return A Token containing the access token and related information
+   * @throws DatabricksException if token exchange fails
```
parthban-db, line 137: ditto: the other IllegalArgumentException error that it can throw.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,200 @@
...
+      throw new DatabricksException(
+          String.format(
+              "Failed to exchange ID token for access token at %s: %s",
+              endpoints.getTokenEndpoint(), e.getMessage()),
+          e);
+    }
+
+    OAuthResponse response;
+    try {
+      response = new ObjectMapper().readValue(rawResponse.getBody(), OAuthResponse.class);
```
parthban-db, line 164: This ObjectMapper is created for each token request. Can we define this as a static final field for this class? Is this readvalue() method thread safe?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,200 @@
...
+import java.time.LocalDateTime;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Implementation of TokenSource that handles OAuth token exchange for Databricks authentication.
+ * This class manages the OAuth token exchange flow using ID tokens to obtain access tokens.
+ */
+public class DatabricksOAuthTokenSource implements TokenSource {
+  /** OAuth client ID used for token exchange */
```
parthban-db, line 19: I didn't find a concrete documentation. However, everywhere in Java docs, they use a full stop to end the sentence for a comment. So, maybe we can follow it.
```suggestion
  /** OAuth client ID used for token exchange. */
```

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,200 @@
...
+    private final String clientId;
+    private final String host;
+    private final OpenIDConnectEndpoints endpoints;
+    private final IDTokenSource idTokenSource;
+    private final HttpClient httpClient;
+    private String accountId;
+    private String audience;
+
+    /**
+     * Validates that a value is non-empty and non-null for required fields.
```
parthban-db, line 55: I think this comment is not correct. We don't have a non-empty check for all the objects. It is only for the Strings. So, we can clarify it here.

renaudhartert-db (approved):
LGTM overall, deferring final approval to @parthban-db and @hectorcast-db.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
hectorcast-db: On the other hand, calls to this method are not frequent, and soon will be done async.
Specially if `readvalue()` is not thread safe, I would not bother.

hectorcast-db (approved):
LGTM bar the JavaDocs note from Parth

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,216 @@
...
+ * Implementation of TokenSource that handles OAuth token exchange for Databricks authentication.
+ * This class manages the OAuth token exchange flow using ID tokens to obtain access tokens.
+ */
+public class DatabricksOAuthTokenSource implements TokenSource {
+  /** OAuth client ID used for token exchange. */
+  private final String clientId;
+  /** Databricks account ID, used as audience if provided. */
+  private final String accountId;
+  /** OpenID Connect endpoints configuration. */
+  private final OpenIDConnectEndpoints endpoints;
```
parthban-db, line 24: Can we pass the endpoint directly here? I remember you saying something about this @hectorcast-db.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
parthban-db: It still does not have the error information.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,216 @@
...
+     * @param httpClient HTTP client for making requests.
+     * @throws IllegalArgumentException if any required parameter is null or empty.
+     */
+    public Builder(
+        String clientId,
+        String host,
+        OpenIDConnectEndpoints endpoints,
+        IDTokenSource idTokenSource,
+        HttpClient httpClient) {
+      validate(clientId, "ClientID");
```
hectorcast-db, line 98: To be consistent, we need to validate this on `Token()`, not when creating the TokenSource.

In other providers, we only check if the fields are set when calling an endpoint/need a token. If we check it earlier, the client construction may fail.

hectorcast-db (changes requested):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
hectorcast-db: Getting the endpoint will fail for some workspaces (OIDC is a user setting). 
In golang, we did not pass the endpoints and load them later.
But we should do instead (I will update Go), is to "try/catch" an error getting the endpoint, and pass a null endpoint if we cannot get it.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSource.java
```diff
@@ -0,0 +1,238 @@
...
+   *
+   * @return A Token containing the access token and related information.
+   * @throws DatabricksException when the token exchange fails.
+   * @throws IllegalArgumentException when there is an error code in the response or when required
+   *     parameters are missing.
+   */
+  @Override
+  public Token getToken() {
+    // Validate all required parameters
+    if (!validate(clientId, "ClientID")
```
hectorcast-db, line 156: It may be more helpful to the user if they see which parameter is missing. Shall we hvae validate itself throw the exception with the field name in it?
Also note that validate takes the field name, but it does not use it.

databricks-sdk-java/src/test/java/com/databricks/sdk/core/oauth/DatabricksOAuthTokenSourceTest.java
```diff
@@ -0,0 +1,296 @@
...
+      throw new RuntimeException("Test failed", e);
+    }
+  }
+
+  /**
+   * Tests validation of required fields in the token source. Verifies that null or empty values for
+   * required fields cause getToken() to throw IllegalArgumentException.
+   */
+  @Test
+  void testParameterValidation() {
```
parthban-db, line 251: Can this be written as a parameterized test?

parthban-db (approved):
LGTM!

hectorcast-db (approved):

PR #444, opened 2025-05-07
Add GitHub OIDC authentication support

emmyzhou-db:
This PR adds support for GitHub OIDC (OpenID Connect) authentication in the Databricks SDK Java.

## Key Changes
- Added `GithubIDTokenSource` to retrieve JWT tokens from GitHub Actions environment
- Introduced new `TokenSourceCredentialsProvider` as a generic provider for token-based authentication flows
- Updated `DefaultCredentialsProvider` to support GitHub OIDC and restructured it to facilitate adding other identity providers
- Added an example demonstrating GitHub OIDC authentication with GitHub Actions

## How is this tested?
- Added unit tests for `GithubIDTokenSource` and `TokenSourceCredentialsProvider`
- Manually validated the authentication flow in a GitHub Actions workflow using the provided example

NO_CHANGELOG=true

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
@@ -100,9 +100,9 @@ public class DefaultCredentialsProvider implements CredentialsProvider {
    * @param config The Databricks configuration containing OIDC settings
    */
   private void addOIDCCredentialsProviders(DatabricksConfig config) {
-    // TODO: refactor the code so that the IdTokenSources are created within the 
+    // TODO: refactor the code so that the IdTokenSources are created within the
     // configure call of their corresponding CredentialsProvider. This will allow
-    // us to simplify the code by validating IdTokenSources when they are created. 
+    // us to simplify the code by validating IdTokenSources when they are created.
     OpenIDConnectEndpoints endpoints = null;
     try {
       endpoints = config.getOidcEndpoints();
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
@@ -100,6 +100,9 @@ public class DefaultCredentialsProvider implements CredentialsProvider {
    * @param config The Databricks configuration containing OIDC settings
    */
   private void addOIDCCredentialsProviders(DatabricksConfig config) {
+    // TODO: refactor the code so that the IdTokenSources are created within the 
+    // configure call of their corresponding CredentialsProvider. This will allow
+    // us to simplify the code by validating IdTokenSources when they are created. 
     OpenIDConnectEndpoints endpoints = null;
     try {
       endpoints = config.getOidcEndpoints();
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
@@ -2,7 +2,6 @@ package com.databricks.sdk.core;
 
 import com.databricks.sdk.core.oauth.*;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -10,48 +9,37 @@ import org.slf4j.LoggerFactory;
 public class DefaultCredentialsProvider implements CredentialsProvider {
   private static final Logger LOG = LoggerFactory.getLogger(DefaultCredentialsProvider.class);
 
-  private static final List<Class<?>> providerClasses =
-      Arrays.asList(
-          PatCredentialsProvider.class,
-          BasicCredentialsProvider.class,
-          OAuthM2MServicePrincipalCredentialsProvider.class,
-          GithubOidcCredentialsProvider.class,
-          AzureGithubOidcCredentialsProvider.class,
-          AzureServicePrincipalCredentialsProvider.class,
-          AzureCliCredentialsProvider.class,
-          ExternalBrowserCredentialsProvider.class,
-          DatabricksCliCredentialsProvider.class,
-          NotebookNativeCredentialsProvider.class,
-          GoogleCredentialsCredentialsProvider.class,
-          GoogleIdCredentialsProvider.class);
-
-  private final List<CredentialsProvider> providers;
+  private List<CredentialsProvider> providers = new ArrayList<>();
 
   private String authType = "default";
+
+  private static class NamedIDTokenSource {
+    private final String name;
+    private final IDTokenSource idTokenSource;
+
+    public NamedIDTokenSource(String name, IDTokenSource idTokenSource) {
+      this.name = name;
+      this.idTokenSource = idTokenSource;
+    }
+
+    public String getName() {
+      return name;
+    }
+
+    public IDTokenSource getIdTokenSource() {
+      return idTokenSource;
+    }
+  }
+
+  public DefaultCredentialsProvider() {}
 
   public String authType() {
     return authType;
   }
 
-  public DefaultCredentialsProvider() {
-    providers = new ArrayList<>();
-    for (Class<?> clazz : providerClasses) {
-      try {
-        providers.add((CredentialsProvider) clazz.newInstance());
-      } catch (NoClassDefFoundError | InstantiationException | IllegalAccessException e) {
-        LOG.warn(
-            "Failed to instantiate credentials provider: "
-                + clazz.getName()
-                + ", skipping. Cause: "
-                + e.getClass().getCanonicalName()
-                + ": "
-                + e.getMessage());
-      }
-    }
-  }
-
   @Override
   public synchronized HeaderFactory configure(DatabricksConfig config) {
+    addDefaultCredentialsProviders(config);
     for (CredentialsProvider provider : providers) {
       if (config.getAuthType() != null
           && !config.getAuthType().isEmpty()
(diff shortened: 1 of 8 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
```diff
@@ -80,4 +93,70 @@ public synchronized HeaderFactory configure(DatabricksConfig config) {
...
+   *
+   * @param config The Databricks configuration containing OIDC settings
+   */
+  private void addOIDCCredentialsProviders(DatabricksConfig config) {
+    OpenIDConnectEndpoints endpoints = null;
+    try {
+      endpoints = config.getOidcEndpoints();
+    } catch (Exception e) {
+      LOG.warn("Failed to get OpenID Connect endpoints", e);
+    }
```
renaudhartert-db, line 108: When I read this code the first time, I got concerned that we might end-up with an NullPointerException downstream. It might be worth throwing an IllegalArgumentException right here so that the intent is clear. This is also less bugprone as it would completely prevent potentials issue if someone change the code to use `endpoints` before the `DatabricksOAuthTokenSource` validation.

```suggestion
    OpenIDConnectEndpoints endpoints;
    try {
      endpoints = config.getOidcEndpoints();
    } catch (Exception e) {
      LOG.warn("Failed to get OpenID Connect endpoints", e);
      throw e;
    }
```

renaudhartert-db (approved):
LGTM overall, deferring approval to @parthban-db

databricks-sdk-java/src/test/java/com/databricks/sdk/core/oauth/GithubIDTokenSourceTest.java
```diff
@@ -0,0 +1,154 @@
...
+                      && request.getUrl().contains("audience=" + TEST_AUDIENCE)
+                      && request
+                          .getHeaders()
+                          .get("Authorization")
+                          .equals("Bearer " + TEST_REQUEST_TOKEN);
+                }));
+  }
+
+  @Test
+  void testSuccessfulTokenRetrievalWithoutAudience() throws IOException {
```
parthban-db, line 73: Can these two tests can be written as parameterized test?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/GithubIDTokenSource.java
```diff
@@ -0,0 +1,110 @@
...
+    if (!jsonResp.has("value")) {
+      throw new DatabricksException("ID token response missing 'value' field");
+    }
+
+    String tokenValue = jsonResp.get("value").textValue();
+    if (Strings.isNullOrEmpty(tokenValue)) {
+      throw new DatabricksException("Received empty ID token from GitHub Actions");
+    }
+
+    return new IDToken(tokenValue);
```
parthban-db, line 108: IDToken constructor can throw IIlegalArgumentException. Since we are not catching it, let's add that in the doccomment.

renaudhartert-db: This will never happen though given the if statement in line 104. 

In almost all cases, we should model optionality using `Optional` and not `null`. It is thus reasonable to assume that a `String` is never null. In that regard, I'm wondering if we should simplify the constructor of the IdToken class so that it does not throw an exception. 

We could do something like this if we really want to be defensive:

```java
public class IDToken {
  // The string value of the ID Token
  private final String value;

  /**
   * Constructs an IDToken with a value.
   *
   * @param value The ID Token string.
   */
  public IDToken(String value) {
    this.value = Strings.isNullOrEmpty(value) ? "" : value;
  }

  /**
   * Returns the value of the ID Token.
   *
   * @return The string representation of the ID Token.
   */
  public String getValue() {
    return value;
  }
}
```

parthban-db: >It would thus be the responsibility of the functions receiving ID tokens to decide whether they are OK with an empty token or not.

Empty Token will always be invalid, right? I can't imagine a scenario where an empty token is valid. So, It would be better to catch at construction. Also, failing fast is a recommended approach in JAVA.

> In almost all cases, it is preferable to model optionality using Optional and not null. That way, we can consider that non-optional objects are never null, except if specifically indicated as nullable.

Agreed, this validation is doing the same by ensuring value can never be null.

renaudhartert-db: > Empty Token will always be invalid, right? I can't imagine a scenario where an empty token is valid.

It's a decision we have to make. I'm personally comfortable with allowing the jwt string to be empty and let the client decide. Though, I'm happy to keep it as-is too.

> failing fast is a recommended approach in JAVA.

Fair but that is true for most things in life, not really specific to this case. One could argue that this is exactly what line 104 is about. 

I don't think listing the runtime exceptions of the underlying functions is a scalable solution though. If we're going to keep the check in the constructor, then I would prefer something like this:

```java
    try {
      String tokenValue = jsonResp.get("value").textValue();
      return new IdToken(tokenValue);
    } catch (IllegalArgumentException e) {
      throw new IllegalArgumentException("...");
    };
```

In the same line of thought, should we move the validation at the beginning of `getIdToken` to the constructor?

emmyzhou-db: Following the previous PR, @hectorcast-db suggested moving the parameter validation from the constructor to the getToken() method for consistency with other providers, which only validate required fields when a token is actually requested. I adapted this change for the DatabricksOAuthTokenSource class, which accepts an IDTokenSource as input. If validation were to fail during construction, it would prevent the instantiation of DatabricksOAuthTokenSource altogether. As this may fail client construction, I decided to defer validation to getIDToken() instead as this would be called inside getToken() of the wrapper class. @hectorcast-db Let me know if my understanding is correct :)

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
emmyzhou-db: From my understanding, we aim to instantiate all providers regardless of whether they are correctly configured, and only fail when a token is actually requested. Throwing an exception during instantiation would prevent OIDC providers from being created. Instead, we could return the exception as you mentioned or, as done in the Go SDK, pass the endpoint getter to the token source. @hectorcast-db suggested that using a try-catch block at this stage might be preferable, but we can discuss the best approach.

hectorcast-db: We definitely not want to throw an error here, because any user which has OIDC disabled in their workspace would be blocked from using the SDKs. We have two options:
* Skip the creation of the CredentialsStrategy
* Create the CredentialsStrategy with a null parameter

I think the second option is more inlined with the rest of the SDKs: We create all CredentialsStrategy even if some parameters are missing and do the validation at `Configure` time. 
This also impacts the logs. When configuring, users will see the `skipping xxx auth` and possible error logs for those which are skipped. If we don't create the CredentialsStrategy at all, the user wont' have the logs (or will have them in a different place if we log it here where the error is thrown).

But I am not strongly  opposed to skip the creation since we are planning to rethink DefaultAuth for SDK Mod.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/GithubIDTokenSource.java
renaudhartert-db: Thanks for the context @emmyzhou-db.

>  If validation were to fail during construction, it would prevent the instantiation of DatabricksOAuthTokenSource altogether. 

Is there a scenario in which it would fail at construction but succeed later?

hectorcast-db: See my message above. Same reasoning applies.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
```diff
@@ -2,56 +2,69 @@
...
-          AzureServicePrincipalCredentialsProvider.class,
-          AzureCliCredentialsProvider.class,
-          ExternalBrowserCredentialsProvider.class,
-          DatabricksCliCredentialsProvider.class,
-          NotebookNativeCredentialsProvider.class,
-          GoogleCredentialsCredentialsProvider.class,
-          GoogleIdCredentialsProvider.class);
-
-  private final List<CredentialsProvider> providers;
+  // List of credential providers that will be tried in sequence
```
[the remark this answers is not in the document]
hectorcast-db: JavaDoc is generally not generated for private elements.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/GithubIDTokenSource.java
renaudhartert-db: I'm not sure I agree with that; `IdTokenSources` are configured at construction.

renaudhartert-db: From offline discussion:

- IdTokenSources should ideally be validated at creation. The reason we cannot do that at the moment is because we are not creating them within the `configure` function of their corresponding `CredentialsProvider`. 
- We will add a TODO in the code to make that clear and address that in a follow-up PR. For now, let's keep validating in the `getIdToken()` method.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
renaudhartert-db: Closing this thread as per the comment.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
```diff
@@ -80,4 +93,71 @@ public synchronized HeaderFactory configure(DatabricksConfig config) {
...
             + " to configure credentials for your preferred authentication method");
   }
+
+  /**
+   * Adds OpenID Connect (OIDC) based credential providers to the list of available providers.
+   *
+   * @param config The Databricks configuration containing OIDC settings
+   */
+  private void addOIDCCredentialsProviders(DatabricksConfig config) {
+    OpenIDConnectEndpoints endpoints = null;
```
renaudhartert-db, line 103:
```suggestion
  private void addOIDCCredentialsProviders(DatabricksConfig config) {
    // TODO: refactor the code so that the IdTokenSources are created within the 
    // configure call of their corresponding CredentialsProvider. This will allow
    // us to simplify the code by validation IdTokenSources when they are created. 
    OpenIDConnectEndpoints endpoints = null;
```

hectorcast-db (approved):

databricks-sdk-java/src/test/java/com/databricks/sdk/DatabricksAuthManualTest.java
```diff
@@ -2,12 +2,17 @@
 
 import com.databricks.sdk.core.ConfigResolving;
 import com.databricks.sdk.core.DatabricksConfig;
+import com.databricks.sdk.core.DummyHttpClient;
```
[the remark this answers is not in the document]
parthban-db: This still uses the dummyHttpClient.

databricks-sdk-java/src/test/java/com/databricks/sdk/DatabricksAuthManualTest.java
```diff
@@ -2,12 +2,17 @@
...
 import com.databricks.sdk.core.utils.TestOSUtils;
 import java.util.Map;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
 public class DatabricksAuthManualTest implements ConfigResolving {
+  private DatabricksConfig createConfigWithMockClient() {
+    return new DatabricksConfig().setHttpClient(new DummyHttpClient());
+  }
```
parthban-db, line 14: ditto it still uses the dummyHttpClient.

parthban-db (approved):

PR #445, opened 2025-05-12
Add support for OIDC ID token authentication using an environment variables and files

emmyzhou-db:
Add Environment Variable and File-based ID Token Sources

## Key Changes
- Added `EnvVarIDTokenSource` to read ID tokens from environment variables
- Added `FileIDTokenSource` to read ID tokens from files
- Both implement the `IDTokenSource` interface for OIDC authentication
- Added envvar-oidc and file-oidc to the list of auth types

## How is this tested?
- Unit tests for both classes

NO_CHANGELOG=true

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
@@ -1,6 +1,7 @@
 package com.databricks.sdk.core;
 
 import com.databricks.sdk.core.oauth.*;
+import com.google.common.base.Strings;
 import java.util.ArrayList;
 import java.util.List;
 import org.slf4j.Logger;
@@ -111,6 +112,20 @@ public class DefaultCredentialsProvider implements CredentialsProvider {
     }
 
     List<NamedIDTokenSource> namedIdTokenSources = new ArrayList<>();
+    namedIdTokenSources.add(
+        new NamedIDTokenSource(
+            "env-oidc",
+            new EnvVarIDTokenSource(
+                // Use configured environment variable name if set, otherwise default to
+                // DATABRICKS_OIDC_TOKEN
+                (Strings.isNullOrEmpty(config.getOidcTokenEnv()))
+                    ? "DATABRICKS_OIDC_TOKEN"
+                    : config.getOidcTokenEnv(),
+                config.getEnv())));
+
+    namedIdTokenSources.add(
+        new NamedIDTokenSource("file-oidc", new FileIDTokenSource(config.getOidcTokenFilepath())));
+
     namedIdTokenSources.add(
         new NamedIDTokenSource(
             "github-oidc",
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
@@ -118,7 +118,7 @@ public class DefaultCredentialsProvider implements CredentialsProvider {
             new EnvVarIDTokenSource(
                 // Use configured environment variable name if set, otherwise default to
                 // DATABRICKS_OIDC_TOKEN
-                (Strings.isNullOrEmpty(config.getOidcTokenEnv()))
+                Strings.isNullOrEmpty(config.getOidcTokenEnv())
                     ? "DATABRICKS_OIDC_TOKEN"
                     : config.getOidcTokenEnv(),
                 config.getEnv())));
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
@@ -32,11 +32,8 @@ public class EnvVarIDTokenSource implements IDTokenSource {
    *
    * @param audience The intended recipient of the ID Token (unused in this implementation).
    * @return An {@link IDToken} containing the token value from the environment variable.
-   * @throws IllegalArgumentException if the environment variable name is null or empty, or the
-   *     environment is null.
+   * @throws IllegalArgumentException if the environment variable name is null or empty.
    * @throws DatabricksException if the environment variable is not set or is empty.
-   * @throws NullPointerException if the environment variable name is null.
-   * @throws ClassCastException if the environment variable name is not a String.
    */
   @Override
   public IDToken getIDToken(String audience) {
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
@@ -6,11 +6,6 @@ import com.google.common.base.Strings;
 
 /** Implementation of {@link IDTokenSource} that reads the ID token from an environment variable. */
 public class EnvVarIDTokenSource implements IDTokenSource {
-  private static final String ERROR_ENV_VAR_NULL_OR_EMPTY =
-      "Environment variable name cannot be null or empty";
-  private static final String ERROR_EMPTY_TOKEN =
-      "Received empty ID token from environment variable %s";
-
   /* The name of the environment variable to read the ID token from. */
   private final String envVarName;
   /* The environment to read variables from. */
@@ -38,14 +33,15 @@ public class EnvVarIDTokenSource implements IDTokenSource {
   @Override
   public IDToken getIDToken(String audience) {
     if (Strings.isNullOrEmpty(envVarName)) {
-      throw new IllegalArgumentException(ERROR_ENV_VAR_NULL_OR_EMPTY);
+      throw new IllegalArgumentException("Environment variable name cannot be null or empty");
     }
 
     try {
       String token = env.get(envVarName);
       return new IDToken(token);
     } catch (IllegalArgumentException e) {
-      throw new DatabricksException(String.format(ERROR_EMPTY_TOKEN, envVarName), e);
+      throw new DatabricksException(
+          String.format("Received empty ID token from environment variable %s", envVarName), e);
     }
   }
 }
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
@@ -41,10 +41,12 @@ public class EnvVarIDTokenSource implements IDTokenSource {
       throw new IllegalArgumentException("Environment cannot be null");
     }
 
-    String token = env.get(envVarName);
-    if (Strings.isNullOrEmpty(token)) {
-      throw new DatabricksException("Missing environment variable: " + envVarName);
+    try {
+      String token = env.get(envVarName);
+      return new IDToken(token);
+    } catch (IllegalArgumentException e) {
+      throw new DatabricksException(
+          "Received empty ID token from environment variable " + envVarName);
     }
-    return new IDToken(token);
   }
 }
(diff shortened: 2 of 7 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/FileIDTokenSource.java
```diff
@@ -0,0 +1,71 @@
...
+   */
+  @Override
+  public IDToken getIDToken(String audience) {
+    if (Strings.isNullOrEmpty(filePath)) {
+      throw new IllegalArgumentException("File path cannot be null or empty");
+    }
+
+    try {
+      Path path = Paths.get(filePath);
+      if (!Files.exists(path)) {
```
parthban-db, line 50: get and exists throw an exception. Either catch it or add it to the doc comment.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/FileIDTokenSource.java
```diff
@@ -0,0 +1,71 @@
...
+        throw new DatabricksException("File " + filePath + " does not exist");
+      }
+
+      List<String> lines = Files.readAllLines(path, StandardCharsets.UTF_8);
```
parthban-db, line 54: ditto readalllines

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/FileIDTokenSource.java
```diff
@@ -0,0 +1,71 @@
...
+      if (lines.isEmpty()) {
+        throw new DatabricksException("File " + filePath + " is empty");
+      }
+
+      try {
+        String token = lines.get(0).trim();
```
parthban-db, line 60: ditto get and trim

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
```diff
@@ -0,0 +1,52 @@
...
+    if (Strings.isNullOrEmpty(envVarName)) {
+      throw new IllegalArgumentException("Environment variable name cannot be null or empty");
+    }
+
+    if (env == null) {
+      throw new IllegalArgumentException("Environment cannot be null");
+    }
+
+    try {
+      String token = env.get(envVarName);
```
parthban-db, line 45: The get method here can also throw an exception, but it is not listed in its doc comment. Can you modify the doccomment for the get method?

emmyzhou-db: I believe Environment.get() implements the Map.get() interface in Java which does not seem to throw an exception, it just returns null if the key is not found. Then the IllegalArgumentException would occur at IDToken construction instead, which is documented and caught.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/FileIDTokenSource.java
emmyzhou-db: Agreed for get(), but is exists() not just a boolean check?

parthban-db: This is what I see from the doc comment of the exists function.

Throws:
SecurityException - In the case of the default provider, the SecurityManager.checkRead(String) is invoked to check read access to the file.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
parthban-db: This is what I see from the doc comment.

Throws:
ClassCastException - if the key is of an inappropriate type for this map (optional)
NullPointerException - if the specified key is null and this map does not permit null keys (optional)

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DefaultCredentialsProvider.java
```diff
@@ -111,6 +112,20 @@ private void addOIDCCredentialsProviders(DatabricksConfig config) {
...
     List<NamedIDTokenSource> namedIdTokenSources = new ArrayList<>();
+    namedIdTokenSources.add(
+        new NamedIDTokenSource(
+            "env-oidc",
+            new EnvVarIDTokenSource(
+                // Use configured environment variable name if set, otherwise default to
+                // DATABRICKS_OIDC_TOKEN
+                (Strings.isNullOrEmpty(config.getOidcTokenEnv()))
+                    ? "DATABRICKS_OIDC_TOKEN"
+                    : config.getOidcTokenEnv(),
```
renaudhartert-db, line 123: Are these parentheses needed?

```suggestion
                Strings.isNullOrEmpty(config.getOidcTokenEnv())
                    ? "DATABRICKS_OIDC_TOKEN"
                    : config.getOidcTokenEnv(),
```

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
```diff
@@ -0,0 +1,56 @@
...
+   */
+  @Override
+  public IDToken getIDToken(String audience) {
+    if (Strings.isNullOrEmpty(envVarName)) {
+      throw new IllegalArgumentException("Environment variable name cannot be null or empty");
+    }
+
+    if (env == null) {
+      throw new IllegalArgumentException("Environment cannot be null");
+    }
```
renaudhartert-db, line 43: Can `config.getEnv()` return a null `Environment`? If not, then let's not test for null here and follow the principle that optionality should be modeled using `Optional` as we discussed in a previous PR.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
renaudhartert-db: Was this resolve? I see the exception is caught.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/FileIDTokenSource.java
```diff
@@ -0,0 +1,105 @@
...
+      lines = Files.readAllLines(path, StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      throw new DatabricksException(
+          "Failed to read ID token from file " + filePath + ": " + e.getMessage(), e);
+    } catch (SecurityException e) {
+      throw new DatabricksException(
+          "Security permission denied when reading file " + filePath + ": " + e.getMessage(), e);
+    }
+
+    if (lines.isEmpty()) {
```
renaudhartert-db, line 89: Should we fail if the file contains more than one non-empty line?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/FileIDTokenSource.java
```diff
@@ -0,0 +1,105 @@
...
+      String token;
+      try {
+        token = lines.get(0).trim();
+      } catch (IndexOutOfBoundsException e) {
+        throw new DatabricksException("Invalid token format in file " + filePath);
+      }
+      return new IDToken(token);
+    } catch (IllegalArgumentException e) {
+      throw new DatabricksException("Received empty ID token from file " + filePath);
+    }
```
renaudhartert-db, line 103: Separate the two tries rather than nesting them. This makes it easier to read and is less bug-prone (e.g. if someone modifies the catch clause of the surrounding try-catch and accidentally catches the exception from the inner throw).

```suggestion
    String token;
    try {
      token = lines.get(0).trim();
    } catch (IndexOutOfBoundsException e) {
      throw new DatabricksException("Invalid token format in file " + filePath);
    }
    
    try {
      return new IDToken(token);
    } catch (IllegalArgumentException e) {
      throw new DatabricksException("Received empty ID token from file " + filePath);
    }
```

renaudhartert-db (changes requested):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
emmyzhou-db: Yes, I checked the docs again and it is true that ClassCastException and NullPointerException can be thrown from Map.get() but in theory this should never happen to keys which are strings and non-null. Should I keep the catch?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
emmyzhou-db: It seems like ```config.getEnv()``` never returns a null ```Environment``` since ```DatabricksConfig``` requires ```env``` to be non-null. I will remove the check.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/EnvVarIDTokenSource.java
parthban-db: Java guidelines recommends these exceptions to propagate as these errors should not occur with how this function is implemented. Still guidelines recommends to simply doc comment this exceptions.

databricks-sdk-java/src/test/java/com/databricks/sdk/core/oauth/FileIDTokenSourceTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("provideTestCases")
+  void testGetIDToken(
+      String testName,
+      String filePath,
+      String fileContent,
+      String expectedToken,
+      Class<? extends Exception> expectedException)
+      throws IOException {
```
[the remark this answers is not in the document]
emmyzhou-db: Agreed, that's a great idea!

databricks-sdk-java/src/test/java/com/databricks/sdk/core/oauth/FileIDTokenSourceTest.java
```diff
@@ -0,0 +1,72 @@
...
+        // expected exception
+        Arguments.of("Valid token file", null, TEST_TOKEN, TEST_TOKEN, null),
+        Arguments.of("Token with whitespace", null, "  " + TEST_TOKEN + "  ", TEST_TOKEN, null),
+        Arguments.of("Empty file", null, "", null, DatabricksException.class),
+        Arguments.of("File with only whitespace", null, "   ", null, DatabricksException.class),
+        Arguments.of("Null file path", null, null, null, IllegalArgumentException.class),
+        Arguments.of("Empty file path", "", null, null, IllegalArgumentException.class),
+        Arguments.of(
+            "Non-existent file",
+            "/path/to/nonexistent/file.txt",
```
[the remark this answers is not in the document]
emmyzhou-db: True, initially, the path was hardcoded because the temp directory is only available at runtime, but I’ve since refactored the test so that the non-existent file path is always resolved within the temp directory.

databricks-sdk-java/src/test/java/com/databricks/sdk/core/oauth/FileIDTokenSourceTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private static Stream<Arguments> provideTestCases() {
+    return Stream.of(
+        // Test case name, file path (null means create temp file), file content, expected token,
+        // expected exception
+        Arguments.of("Valid token file", null, TEST_TOKEN, TEST_TOKEN, null),
+        Arguments.of("Token with whitespace", null, "  " + TEST_TOKEN + "  ", TEST_TOKEN, null),
+        Arguments.of("Empty file", null, "", null, DatabricksException.class),
+        Arguments.of("File with only whitespace", null, "   ", null, DatabricksException.class),
+        Arguments.of("Null file path", null, null, null, IllegalArgumentException.class),
```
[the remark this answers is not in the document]
emmyzhou-db: The file is only created if the file content is not null, so in cases where the content is null, no file will be created. But I agree that the previous logic was a bit convoluted, so I’ve refactored the tests to make them more straightforward and intuitive.

renaudhartert-db (approved):
LGTM! Thanks Emmy

parthban-db (approved):

PR #453, opened 2025-05-26
Add support for direct dataplane access

emmyzhou-db:
This PR introduces direct dataplane access to the Java SDK.

### 1. OAuth Header Factory Implementation
- Introduced `OAuthHeaderFactory` interface that combines `HeaderFactory` and `TokenSource` functionality
### 2. Error Token Source
- Added `ErrorTokenSource` implementation that provides clear error messages when OAuth tokens are not supported

### 3. OAuth Authentication Flow Changes
- Modified OAuth auth types to return `OAuthHeaderFactory` instead of basic `HeaderFactory`
- Updated credential providers to use the new OAuth header factory system:
  - Azure Service Principal
  - OAuth M2M Service Principal
  - External Browser
  - Databricks CLI
  - GitHub OIDC
  - Azure GitHub OIDC

## How is this tested?
Added unit tests for all new classes including `OAuthHeaderFactory`, `ErrorTokenSource`, along with tests for `config.getTokenSource()`. The feature has been manually validated end-to-end by creating a model serving endpoint with route optimization enabled and successfully sending queries through the dataplane access path.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/service/serving/ServingEndpointsDataPlaneImpl.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/service/serving/ServingEndpointsDataPlaneImpl.java
@@ -48,16 +48,13 @@ class ServingEndpointsDataPlaneImpl implements ServingEndpointsDataPlaneService 
   public QueryEndpointResponse query(QueryEndpointInput request) {
     DataPlaneInfo dataPlaneInfo = dataPlaneInfoQuery(request);
     String path = dataPlaneInfo.getEndpointUrl();
-    Token token =
-        dataPlaneTokenSource.getToken(
-            dataPlaneInfo.getEndpointUrl(), dataPlaneInfo.getAuthorizationDetails());
+    Token token = dataPlaneTokenSource.getToken(path, dataPlaneInfo.getAuthorizationDetails());
 
     try {
       Request req = new Request("POST", path, apiClient.serialize(request));
       ApiClient.setQuery(req, request);
       req.withHeader("Accept", "application/json");
       req.withHeader("Content-Type", "application/json");
-
       RequestOptions options =
           new RequestOptions()
               .withAuthorization(token.getTokenType() + " " + token.getAccessToken())


databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -195,11 +195,32 @@ public <T> T execute(Request in, Class<T> target) throws IOException {
...
+    return deserialize(out, target);
+  }
+
   private Response getResponse(Request in) {
-    return executeInner(in, in.getUrl());
+    return executeInner(in, in.getUrl(), Optional.empty());
+  }
+
+  private Response getResponse(Request in, RequestOptions options) {
+    return executeInner(in, in.getUrl(), Optional.of(options));
```
parthban-db, line 220: I think we don't need these two methods as they are private.

renaudhartert-db:
Thanks @emmyzhou-db! This is a big PR and will take a little time to review it properly. I think this is actually a good example of a PR that could have been broken down into smaller atomic PRs. For example, we could have had a first PR focused only on adding the new header factory and changing existing credentials providers.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -195,11 +195,32 @@ public <T> T execute(Request in, Class<T> target) throws IOException {
...
+  /**
+   * Executes HTTP request with retries and converts it to proper POJO, using custom request
+   * options.
+   *
+   * @param in Commons HTTP request
+   * @param target Expected pojo type
+   * @param options Optional request options to customize request behavior
+   * @return POJO of requested type
+   */
+  public <T> T execute(Request in, Class<T> target, RequestOptions options) throws IOException {
```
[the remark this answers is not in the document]
emmyzhou-db: That's cleaner indeed. Done!

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
emmyzhou-db: That's true. This would work in this case since a default `RequestOptions` applies the identity function to a request, but then there is no need for an optional since `options` will always be present anyways, just doing nothing to the request depending on which `execute()` was called.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/ApiClient.java
```diff
@@ -173,33 +173,46 @@ public Map<String, String> getStringMap(Request req) {
...
+    Response out = getResponse(in, options);
     if (target == Void.class) {
       return null;
     }
     return deserialize(out, target);
   }
 
-  private Response getResponse(Request in) {
-    return executeInner(in, in.getUrl());
+  private Response getResponse(Request in, RequestOptions options) {
```
renaudhartert-db, line 211: Could we merge this with `executeInner`?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -209,6 +212,23 @@ public synchronized Map<String, String> authenticate() throws DatabricksExceptio
...
+        setAuthType(credentialsProvider.authType());
+      }
+      if (headerFactory instanceof OAuthHeaderFactory) {
+        return (TokenSource) headerFactory;
+      }
+      return new ErrorTokenSource(
+          String.format("OAuth Token not supported for current auth type %s", authType));
+    } catch (Exception e) {
+      return new ErrorTokenSource("Failed to get token source: " + e.getMessage());
+    }
```
renaudhartert-db, line 229: Restrict the scope of the try catch to the specific instruction that might throw an exception. IIUC, this would be `configure`?

```suggestion
      if (headerFactory == null) {
        try {
          ConfigLoader.fixHostIfNeeded(this);
          headerFactory = credentialsProvider.configure(this);
        } catch (Exception e) {
          return new ErrorTokenSource("Failed to get token source: " + e.getMessage());
        }     
        setAuthType(credentialsProvider.authType());
      }
      
      if (headerFactory instanceof OAuthHeaderFactory) {
        return (TokenSource) headerFactory;
      }
      return new ErrorTokenSource(
          String.format("OAuth Token not supported for current auth type %s", authType));
```

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/OAuthHeaderFactoryFromSuppliers.java
```diff
@@ -0,0 +1,35 @@
+package com.databricks.sdk.core.oauth;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+/**
+ * Implementation of {@link OAuthHeaderFactory} that uses separate suppliers for token and header
+ * generation.
+ */
+public class OAuthHeaderFactoryFromSuppliers implements OAuthHeaderFactory {
```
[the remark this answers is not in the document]
emmyzhou-db: That makes sense, thanks! I've adapted the second approach and removed the `OAuthHeaderFactoryFromTokenSource` and `OAuthHeaderFactoryFromSuppliers` classes entirely.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
emmyzhou-db: FIxed!

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/OAuthHeaderFactoryFromSuppliers.java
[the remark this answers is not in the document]
emmyzhou-db: This class has been removed now :)

databricks-sdk-java/src/main/java/com/databricks/sdk/service/serving/ServingEndpointsDataPlaneAPI.java
```diff
@@ -17,8 +18,9 @@ public class ServingEndpointsDataPlaneAPI {
   private final ServingEndpointsDataPlaneService impl;
 
   /** Regular-use constructor */
-  public ServingEndpointsDataPlaneAPI(ApiClient apiClient) {
-    impl = new ServingEndpointsDataPlaneImpl(apiClient);
+  public ServingEndpointsDataPlaneAPI(
+      ApiClient apiClient, DatabricksConfig config, ServingEndpointsAPI servingEndpointsAPI) {
+    impl = new ServingEndpointsDataPlaneImpl(apiClient, config, servingEndpointsAPI);
```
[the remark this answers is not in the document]
emmyzhou-db: I am not sure how the procedure for syncing the generated code and hand-written code works, but I have made a PR on universe for the updated templates. Perhaps you can take a look at it to make sure everything is fine and then I will remove it from the PR here along with `WorkspaceClient` and the implementation :)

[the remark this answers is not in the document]
renaudhartert-db: SG, let's merge the codegen first 👍

renaudhartert-db (approved):
LGTM, to be merged once we have merged the code generation PR first.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -209,6 +212,24 @@ public synchronized Map<String, String> authenticate() throws DatabricksExceptio
     }
   }
 
+  public TokenSource getTokenSource() {
+    if (headerFactory == null) {
+      try {
+        ConfigLoader.fixHostIfNeeded(this);
+        headerFactory = credentialsProvider.configure(this);
+      } catch (Exception e) {
+        return new ErrorTokenSource("Failed to get token source: " + e.getMessage());
```
hectorcast-db, line 221: Why do we do this instead of throwing an exception?

emmyzhou-db: This is adopted from how it is done in the Go SDK. Essentially, this method is called during workspace client construction and we don't want that to fail so we wrap it with an `ErrorTokenSource`. That way it will only throw an exception later when the actual data plane API call is being made.

parthban-db (approved):

PR #455, opened 2025-05-30
Add Asynchronous Token Refresh to `RefreshableTokenSource`

emmyzhou-db:
## Motivation
Synchronous token refresh introduces unnecessary blocking and latency, which becomes especially problematic in high QPS (Queries Per Second) workloads. In such scenarios, blocking token refresh operations can create bottlenecks, impact throughput, and cause request queuing.

## Solution
This PR introduces asynchronous token refresh to improve responsiveness and reliability, with particular benefits for high-throughput applications.

 **Async Refresh Option**: Added `withAsyncRefresh(boolean enabled)` method to enable/disable asynchronous token refresh. When enabled, tokens are refreshed in the background when they become "stale" (close to expiry).

### Token State Management
- **Three-State Token System**:
  - `FRESH`: Token is valid and not close to expiry
  - `STALE`: Token is valid but will expire soon (triggers async refresh if enabled)
  - `EXPIRED`: Token has expired (requires blocking refresh)
 - `Token` class is now a pure data class, holding only token information (access token, refresh token, expiry time).
 - `TokenSource` implementations are responsible for token state management and refresh logic.

### Performance Optimizations
- **Non-blocking for Stale Tokens**: When async is enabled and the token is stale but not expired, API calls continue using the current token while a background refresh is triggered, reducing latency and avoiding thread blocking.
- **Blocking Only on Expiry**: If the token is expired, calls will still block until a new token is fetched, ensuring correctness.
- **Default Stale Duration**: Tokens are considered stale 3 minutes before expiry, allowing proactive refresh while maintaining validity.

### Thread Safety & Reliability
- **Synchronized Refresh Logic**: All async refresh operations are synchronized to prevent race conditions and redundant refreshes.
- **Refresh State Tracking**: The implementation tracks:
  - Whether a refresh is already in progress
  - Whether the last refresh succeeded
  - This prevents unnecessary or repeated background refreshes
- **Failure Handling**: If an async refresh fails, subsequent refreshes will be forced to be synchronous until a successful refresh occurs.

### Configuration Options
**Experimental Features (may change in future releases)**:
- withClockSupplier() - Custom clock supplier for testing
- withAsyncRefresh() - Enable/disable asynchronous token refresh
- withExpiryBuffer() - Configure the expiry buffer duration

## Testing
- **Unit tests** cover async refresh triggering, correct token usage during background refresh, and thread safety.
- **Time Control for Testing:**  
  The `RefreshableTokenSource` class can be configured with a custom clock supplier via `withClockSupplier(ClockSupplier clockSupplier)`, allowing tests to precisely control and simulate token expiry and refresh timing. This enables deterministic testing of token state transitions and refresh behavior without relying on real system time.

NO_CHANGELOG=true


databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -5,37 +5,223 @@
...
+  /** Whether asynchronous refresh is enabled. */
+  private boolean asyncEnabled = false;
+  /** Duration before expiry to consider a token as 'stale'. */
+  private Duration staleDuration = DEFAULT_STALE_DURATION;
+  /** Additional buffer before expiry to consider a token as expired. */
+  private Duration expiryBuffer = Duration.ZERO;
+  /** Whether a refresh is currently in progress (for async refresh). */
+  private boolean refreshInProgress = false;
+  /** Whether the last refresh attempt succeeded. */
+  private boolean lastRefreshSucceeded = true;
```
renaudhartert-db, line 50: [nit] Prefer `//` for non-javadoc comments.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -5,37 +5,223 @@
...
 
+  /** Default constructor. */
```
renaudhartert-db, line 52: What's the behavior of the default constructor?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -5,37 +5,223 @@
...
+  /**
+   * Enable or disable asynchronous token refresh.
+   *
+   * @param enabled true to enable async refresh, false to disable
+   * @return this instance for chaining
+   */
+  public RefreshableTokenSource withAsyncRefresh(boolean enabled) {
+    this.asyncEnabled = enabled;
+    return this;
+  }
```
renaudhartert-db, line 73: Could we add these `withX` building utilities to an option or builder class instead? This would better separate the "construction" from the "usage" concern while also guaranteeing that the class behavior is fixed at construction rather than mutable during its lifespan.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -5,37 +5,223 @@
...
+  private static final Duration DEFAULT_STALE_DURATION = Duration.ofMinutes(3);
+
+  /** The current OAuth token. May be null if not yet fetched. */
   protected Token token;
+  /** Whether asynchronous refresh is enabled. */
+  private boolean asyncEnabled = false;
+  /** Duration before expiry to consider a token as 'stale'. */
+  private Duration staleDuration = DEFAULT_STALE_DURATION;
+  /** Additional buffer before expiry to consider a token as expired. */
+  private Duration expiryBuffer = Duration.ZERO;
```
renaudhartert-db, line 46: Should this be initialized to 40 seconds?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -5,37 +5,223 @@
...
+    }
+    try {
+      Token newToken = refresh();
+      token = newToken;
+      lastRefreshSucceeded = true;
+      return newToken;
+    } catch (Exception e) {
+      lastRefreshSucceeded = false;
+      throw e;
+    }
```
renaudhartert-db, line 154: Keep the `try` statement focused on the instruction that throws.

```java
    lastRefreshSucceeded = false;
    try {
      token = refresh();
    } catch (Exception e) {
      throw e;
    }
    lastRefreshSucceeded = true;
    return token;
```

Doing the above makes it clear that the `try ... catch` block does do much as it is equivalent to:

```java
    lastRefreshSucceeded = false;
    token = refresh(); // may throw
    lastRefreshSucceeded = true;
    return token;
```

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -5,37 +5,223 @@
...
+      state = getTokenState();
+      currentToken = token;
+    }
+    if (state == TokenState.FRESH) {
+      return currentToken;
+    }
+    if (state == TokenState.STALE) {
+      // Trigger background refresh, return current token
+      triggerAsyncRefresh();
+      return token;
```
renaudhartert-db, line 179: Should this be `currentToken`?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -5,37 +5,223 @@
...
+    } else {
+      // Token is expired, block to refresh
+      return getTokenBlocking();
+    }
```
renaudhartert-db, line 183: Prefer an exhaustive switch statement. These are easier to read, make it clear that the cases are exclusive, and is also less bug-prone (e.g. if a new state is added). 

```java
switch (state) {
    case FRESH:
        return currentToken;
    case STALE:
        triggerAsyncRefresh(); // trigger background refresh
        return currentToken;
    case EXPIRED:
        return getTokenBlocking();
    default:
        throw new IllegalStateException("Unexpected TokenState: " + state);
}
```

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -5,37 +5,223 @@
...
+  /**
+   * Gets the current token, refreshing if necessary. If async refresh is enabled, may return a
+   * stale token while a refresh is in progress.
+   *
+   * <p>This method may throw an exception if the token cannot be refreshed, depending on the
+   * implementation of {@link #refresh()}.
+   *
+   * @return The current valid token
+   */
+  public synchronized Token getToken() {
```
parthban-db, line 106: This will synchronize every thread that is trying to get a token.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -5,37 +5,223 @@
...
+    EXPIRED
+  }
+
+  // Default duration before expiry to consider a token as 'stale'.
+  private static final Duration DEFAULT_STALE_DURATION = Duration.ofMinutes(3);
+
+  /** The current OAuth token. May be null if not yet fetched. */
   protected Token token;
+  /** Whether asynchronous refresh is enabled. */
+  private boolean asyncEnabled = false;
```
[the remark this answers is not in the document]
renaudhartert-db: It makes sense for the value to originally come from some global context (e.g. an env variable) but I would strongly recommend to keep the cache's behavior (i) self-contained, and (ii) immutable after creation.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
emmyzhou-db: Thanks for your comment! In terms of implementation effort, I think adding an options class would be the quickest solution, since introducing a builder pattern would require updating every single refreshable token source instance. However, because Java doesn’t support constructor inheritance, we’d still need to manually add a constructor that accepts Options to each child class. Given this, do you think it’s better to stick with inheritance, or would it make more sense to use composition instead?

databricks-sdk-java/src/test/java/com/databricks/sdk/core/oauth/RefreshableTokenSourceTest.java
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  private static Stream<Arguments> provideAsyncRefreshScenarios() {
+    return Stream.of(
+        Arguments.of("Fresh token, async enabled", 10, true, false, false),
+        Arguments.of("Stale token, async enabled", 1, true, true, true),
+        Arguments.of("Expired token, async enabled", -1, true, true, true),
+        Arguments.of("Fresh token, async disabled", 10, false, false, false),
+        Arguments.of("Stale token, async disabled", 1, false, false, false),
+        Arguments.of("Expired token, async disabled", -1, false, true, true));
```
renaudhartert-db, line 85: Is there a reason for some expiry to be `1` and other `10`? This is not clear to me.

databricks-sdk-java/src/test/java/com/databricks/sdk/core/oauth/RefreshableTokenSourceTest.java
```diff
@@ -0,0 +1,167 @@
...
+  }
```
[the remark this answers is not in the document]
emmyzhou-db: Yes exactly, so this test is testing whether calling `getToken()` triggers a refresh on different states of the token. If a token is fresh, we do not expect `refresh() `to be called, and if it is stale and async is enabled then we do expect a refresh. So what we are testing is `getToken()` and whether that triggers the refresh as expected, and this is almost instant :)

databricks-sdk-java/src/test/java/com/databricks/sdk/core/oauth/RefreshableTokenSourceTest.java
```diff
@@ -0,0 +1,167 @@
...
+  private static final ZoneId ZONE_ID = ZoneId.of("UTC");
+
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("provideAsyncRefreshScenarios")
+  void testAsyncRefreshParametrized(
+      String testName,
+      long minutesUntilExpiry,
+      boolean asyncEnabled,
+      boolean expectRefresh,
+      boolean expectRefreshedToken)
```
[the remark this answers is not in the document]
emmyzhou-db: This distinction is needed for cases where a refresh is expected to be triggered, but we don’t expect the new token to be returned just yet. Specifically, when the token is stale and async is enabled, we return the current (stale) token—so expectRefreshedToken is false, but expectRefresh is true since `refresh()` should still be called. The previous test made this unclear, so I’ve simplified it for clarity.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/RefreshableTokenSource.java
```diff
@@ -5,36 +5,252 @@
...
+  // Duration before expiry to consider a token as 'stale'.
+  private Duration staleDuration = DEFAULT_STALE_DURATION;
+  // Additional buffer before expiry to consider a token as expired.
+  private Duration expiryBuffer = Duration.ofSeconds(40);
+  // Whether a refresh is currently in progress (for async refresh).
+  private boolean refreshInProgress = false;
+  // Whether the last refresh attempt succeeded.
+  private boolean lastRefreshSucceeded = true;
+  // Clock supplier for current time, for testing purposes.
+  private ClockSupplier clockSupplier = new SystemClockSupplier();
```
parthban-db, line 56: Shouldn't the tests fail with this? As we have changed everything to UTC.

Also, I don't like the idea of having testing code in the main code.

parthban-db: Can you change the name of this class?

parthban-db (approved):
LGTM modulo comments.

renaudhartert-db (approved):

PR #456, opened 2025-06-04
Convert All Usage of `LocalDateTime` to `Instant`

emmyzhou-db:

This pull request refactors the Databricks Java SDK to replace all usages of LocalDateTime with Instant for representing date-time values. The primary motivation is to ensure that all date-time handling is unambiguous, always in UTC, and consistent across the SDK.

## How does time currently operate in the SDK?
Currently, tokens in the SDK use LocalDateTime for expiry timestamps, which by itself is timezone-agnostic and could be ambiguous. Thus, it is not clear how time is handled from the type alone and we need to look at the implementation details of the ClockSupplier used.

The SDK currently uses system time through the `SystemClockSupplier` class as defined [here](https://github.com/databricks/databricks-sdk-java/blob/main/databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/SystemClockSupplier.java). This implementation of the ClockSupplier interface returns `Clock.systemDefaultZone()`, meaning it uses the actual system time from the operating system. The `SystemClockSupplier` is the used throughout the SDK to conduct OAuth token expiration checks as seen [here](https://github.com/databricks/databricks-sdk-java/blob/main/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java) in the `isExpired()` method of the Token class. While `LocalDateTime` doesn't explicitly store timezone information, the system's timezone is being implicitly used for these comparisons through `Clock.systemDefaultZone()`. In other words, token expiry times are implicitly measured in the system's timezone.

## Converting to UTC
To convert these expiry times to UTC while maintaining the SDK's current behavior, we need to handle two distinct scenarios:

### 1. Tokens with `expires_in`

- The response includes the time until expiry.
- **Conversion:**  
  Add the `expires_in` value to the current UTC time.
- **Result:**  
  This approach maintains the original expiration behavior while using UTC internally.

---

#### With Time Zone Indicator

- Tokens (such as those from the Databricks CLI) provide expiry times in RFC 3339 format, including a time zone indicator.
- **Conversion:**  
  These can be easily converted to UTC using standard library methods.

#### Without Time Zone Indicator

- Example: The `expiresOn` field from Azure CLI  
  [Reference](https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/_profile.py#L384)
- Tokens provide expiry timestamps without explicit time zone information.
- The timestamp is implicitly in the system’s local time zone.
- **Conversion steps:**
  1. Interpret the timestamp as local time.
  2. Add the system timezone.
  3. Convert this to UTC.
  We do this with **`LocalDateTime.atZone(systemDefault).toInstant()`**

## How is this tested?
- Existing unit tests to ensure no regressions.
- Added new tests in https://github.com/databricks/databricks-sdk-java/pull/462 which verifies that this change does not change overall behavior of CilTokenSource.

## Future Work
- The SDK currently parses expiry strings from external sources like Azure CLI
- Azure CLI provides two timestamp fields:
  - `expiresOn`: Local datetime without timezone (maintained for backward compatibility)
  - `expires_on`: POSIX timestamp (seconds since epoch) - preferred format

### Current Implementation
- The Java SDK currently uses the deprecated `expiresOn` field
- `LocalDateTime` creates timezone interpretation challenges

### To Do
1. **Phase out `expiresOn`**  
   This legacy field should be deprecated in favor of `expires_on`
2. **Update parsing logic**  
   Future PRs should prioritize using `expires_on` (POSIX timestamp) which:
   - Provides unambiguous UTC-based expiration time
   - Avoids timezone conversion complexities

NO_CHANGELOG=true


databricks-sdk-java/src/main/java/com/databricks/sdk/core/CliTokenSource.java
```diff
@@ -36,21 +38,40 @@ public CliTokenSource(
...
+    }
+
     String multiplePrecisionPattern =
         "[SSSSSSSSS][SSSSSSSS][SSSSSSS][SSSSSS][SSSSS][SSSS][SSS][SS][S]";
     List<String> datePatterns =
         Arrays.asList(
             "yyyy-MM-dd HH:mm:ss",
             "yyyy-MM-dd HH:mm:ss." + multiplePrecisionPattern,
-            "yyyy-MM-dd'T'HH:mm:ss." + multiplePrecisionPattern + "XXX",
-            "yyyy-MM-dd'T'HH:mm:ss." + multiplePrecisionPattern + "'Z'");
```
parthban-db, line 47: Just for my understanding, why did you remove this `"yyyy-MM-dd'T'HH:mm:ss." + multiplePrecisionPattern + "'Z'"`?

emmyzhou-db: Because 'Instant.parse(expiry)' in line 58 parses that format.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/CliTokenSource.java
```diff
@@ -36,21 +38,40 @@ public CliTokenSource(
...
+   * input is ignored.
+   *
+   * <p>The method attempts to parse the input using several common date-time formats, including
+   * ISO-8601 and patterns with varying sub-second precision.
+   *
+   * @param expiry the expiry time string to parse, which must represent a UTC time
+   * @return the parsed {@link Instant}
+   * @throws DateTimeParseException if the input string cannot be parsed
+   */
+  static Instant parseExpiry(String expiry) {
```
renaudhartert-db, line 54: Could we simply use `Instant.parse`? https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#parse-java.lang.CharSequence-

renaudhartert-db: Rephrasing: could you explain why the fallback is necessary?

databricks-sdk-java/src/test/resources/testdata/az
```diff
@@ -22,14 +22,16 @@ for arg in "$@"; do
     fi
 done
 
-# Macos
-EXP="$(/bin/date -v+${EXPIRE:=10S} +'%F %T' 2>/dev/null)"
+# MacOS
```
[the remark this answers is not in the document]
parthban-db: This looks like a mock of Azure CLI, not the Databricks CLI. Does it mean that CLITokenSource is also parsing tokens generated from Azure CLI? Can they use the previous format with a space separator?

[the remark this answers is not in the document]
parthban-db: Can we have this change in another PR? So, this PR will only be a refactor without any behavioral change.

parthban-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/CliTokenSource.java
```diff
@@ -36,21 +39,31 @@ public CliTokenSource(
...
-            "yyyy-MM-dd'T'HH:mm:ss." + multiplePrecisionPattern + "XXX",
-            "yyyy-MM-dd'T'HH:mm:ss." + multiplePrecisionPattern + "'Z'");
-    DateTimeParseException lastException = null;
+        Arrays.asList("yyyy-MM-dd HH:mm:ss", "yyyy-MM-dd HH:mm:ss." + multiplePrecisionPattern);
     for (String pattern : datePatterns) {
       try {
         DateTimeFormatter formatter = DateTimeFormatter.ofPattern(pattern);
         LocalDateTime dateTime = LocalDateTime.parse(expiry, formatter);
-        return dateTime;
+        return dateTime.atZone(ZoneId.systemDefault()).toInstant();
```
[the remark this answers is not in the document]
emmyzhou-db: Updated PR description!

PR #461, opened 2025-06-10
Refactor Token Caching with CachedTokenSource Wrapper

emmyzhou-db:
## Changes
This PR refactors the token refresh and caching in the Java SDK by:
- Removing the `RefreshableTokenSource` class that previously handled token refresh and caching logic directly
- Introducing `CachedTokenSource` as a generic wrapper that can be applied to any TokenSource implementation
- Separating concerns by moving caching and async logic out of individual token sources into a dedicated cache wrapper

## Motivation
Previously, token sources inherited from `RefreshableTokenSource` to gain caching and asynchronous refresh capabilities. However, this inheritance-based approach had limitations when we wanted to make configuration immutable after construction using the builder pattern.

**The Problem**: With abstract classes in Java, constructors are not inherited. To support builder-based configuration, each token source subclass would need to be modified to accept a builder instead of individual parameters, requiring changes across many classes.

**The Solution**: Switch from inheritance to composition. Instead of inheriting caching behavior, we now wrap any `TokenSource` which previously inherited from `RefreshableTokenSource` with `CachedTokenSource` to add caching functionality externally. This allows us to use the builder pattern to configure optional parameters like async refresh and stale duration on each wrapped instance.

## How is this tested?
This change has been tested with async enabled/disabled by default on existing unit and integration tests.

NO_CHANGELOG=true


renaudhartert-db (approved):

databricks-sdk-java/src/test/java/com/databricks/sdk/core/CliTokenSourceTest.java
```diff
@@ -8,6 +8,7 @@
 import static org.mockito.Mockito.mockStatic;
 import static org.mockito.Mockito.when;
 
+import com.databricks.sdk.core.oauth.CachedTokenSource;
```
parthban-db, line 11: IMO, we don't need to wrap the token source in the unit tests. As we are not testing the cliTokenSource class in isolation.

parthban-db (approved):

PR #462, opened 2025-06-12
Add Additional Tests for CliTokenSource

emmyzhou-db:
This PR adds additional tests for the `CliTokenSource` class, specifically focusing on token expiration handling across different timezones and date formats.

## Key Changes
This test suite verifies the behavior of the parseExpiry() method in CliTokenSource by systematically testing tokens with varying expiration times (5, 30, 60, and 120 minutes before and after current system time) across different timezones (UTC, GMT+1, GMT+8, GMT-1, GMT-8). For each combination, it creates a mock token with the specified expiry time, temporarily sets the system timezone, and verifies that the token's expired status matches the expected value regardless of the timezone or date format used.

**Important Note:** This test assumes that all expiry strings provided to `parseExpiry()` are in the system's local timezone. It does not cover scenarios where the expiry string is in a different timezone than where the SDK is running, as this would require additional timezone conversion logic that is not part of the current implementation.

## How is this tested?
N/A

### DIFF
--- a/databricks-sdk-java/src/test/java/com/databricks/sdk/core/CliTokenSourceTest.java
+++ b/databricks-sdk-java/src/test/java/com/databricks/sdk/core/CliTokenSourceTest.java
@@ -9,8 +9,8 @@ import static org.mockito.Mockito.when;
 
 import com.databricks.sdk.core.oauth.Token;
 import com.databricks.sdk.core.utils.Environment;
-import com.databricks.sdk.core.utils.OSUtils;
 import com.databricks.sdk.core.utils.OSUtilities;
+import com.databricks.sdk.core.utils.OSUtils;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.time.Duration;
@@ -22,6 +22,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TimeZone;
 import java.util.stream.Stream;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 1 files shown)


databricks-sdk-java/src/test/java/com/databricks/sdk/core/CliTokenSourceTest.java
```diff
@@ -1,12 +1,154 @@
...
 public class CliTokenSourceTest {
+  private String[] getDateFormats() {
+    return new String[] {
+      "yyyy-MM-dd HH:mm:ss",
+      "yyyy-MM-dd HH:mm:ss.SSS",
+      TimeZone.getDefault().getID().equals("UTC")
+          ? "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
+          : "yyyy-MM-dd'T'HH:mm:ss.SSSXXX"
+    };
+  }
```
parthban-db, line 45: Any reason why these date formats are not in the stream of test cases?

emmyzhou-db: The last element of the array changes depending on the temporary timezone set for the current test running. So in the current setup it would be wrong to have it in the stream of test cases as that is static and evaluated at compile-time. But I can add it as an additional parameter and match it using the timezone, that way it can be part of the stream as well.

databricks-sdk-java/src/test/java/com/databricks/sdk/core/CliTokenSourceTest.java
```diff
@@ -1,12 +1,160 @@
...
+                  new ArrayList<>(
+                      Arrays.asList(
+                          "yyyy-MM-dd HH:mm:ss",
+                          "yyyy-MM-dd HH:mm:ss.SSS",
+                          "yyyy-MM-dd'T'HH:mm:ss.SSSXXX"));
+
+              if (timezone.equals("GMT")) {
+                // We only test with this format when timezone is GMT
+                dateFormats.add("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+              }
```
parthban-db, line 78: Could you add a comment that the Databricks CLI will only return the time in this format when the system is in UTC?

parthban-db (approved):

databricks-sdk-java/src/test/java/com/databricks/sdk/core/CliTokenSourceTest.java
```diff
@@ -1,12 +1,163 @@
...
+                          "yyyy-MM-dd HH:mm:ss",
+                          "yyyy-MM-dd HH:mm:ss.SSS",
+                          "yyyy-MM-dd'T'HH:mm:ss.SSSXXX"));
+
+              if (timezone.equals("GMT")) {
+                /*
+                 * The Databricks CLI outputs timestamps with 'Z' suffix (e.g.,
+                 * 2024-03-14T10:30:00.000Z) only when in UTC/GMT+0 timezone.
+                 * Thus, we only test with this format together with the GMT timezone.
+                 */
```
renaudhartert-db, line 79:
```suggestion
                // The Databricks CLI outputs timestamps with 'Z' suffix (e.g.,
                // 2024-03-14T10:30:00.000Z) only when in UTC/GMT+0 timezone.
                // Thus, we only test with this format together with the GMT timezone.
```

renaudhartert-db (approved):

PR #467, opened 2025-06-19
Custom serializers for Google Protobuf Types

hectorcast-db:
Custom serializers for Google Protobuf Types. This is required because Jackson by default does not consider that such objects have a special serialization.

## How is this tested?
Tested as part of supporting well known types
https://github.com/databricks/databricks-sdk-java/pull/459/files#diff-c3aa175ce608f5cd289331f74de4e21ddcd56fc195ec92b028f2e7d4a6d721e6

NO_CHANGELOG=true


parthban-db (commented):
Can we add unit tests for these classes?

hectorcast-db:
jenkins merge

PR #473, opened 2025-06-22
Remove redundant summary from Java doc.

renaudhartert-db:

This PR regenerates the Java Doc with the new version of the code generator. Specifically, it removes the "summary" section which was, more often than not, redundant with the main documentation. 

## How is this tested?

N/A


databricks-sdk-java/src/main/java/com/databricks/sdk/service/aibuilder/CustomLlmsAPI.java
```diff
@@ -27,7 +27,6 @@ public void cancel(String id) {
     cancel(new CancelCustomLlmOptimizationRunRequest().setId(id));
   }
 
-  /** Cancel a Custom LLM Optimization Run. */
```
parthban-db, line 30: Can you also update the OpenAPI Spec? The summary and description of this old OpenAPI are not compatible with the new genkit.

PR #482, opened 2025-07-11
infer azure tenant id

sreekanth-db:

This PR modifies Azure CLI and Azure SP credential providers to attempt to load the tenant ID of the workspace if not provided before authenticating. Tenant ID is indirectly exposed via the redirect URL used when logging into a workspace. In this PR, we fetch the tenant ID from this endpoint and configure it if not already set.

Reference PR: https://github.com/databricks/databricks-sdk-py/pull/638

## How is this tested?
- Tested the changes using unit tests

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
@@ -14,10 +14,15 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Field;
+import java.net.URL;
 import java.util.*;
 import org.apache.http.HttpMessage;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class DatabricksConfig {
+
+  private static final Logger logger = LoggerFactory.getLogger(DatabricksConfig.class);
   private CredentialsProvider credentialsProvider = new DefaultCredentialsProvider();
 
   @ConfigAttribute(env = "DATABRICKS_HOST")
@@ -414,13 +419,17 @@ public class DatabricksConfig {
     return this;
   }
 
-  /** @deprecated Use {@link #getAzureUseMsi()} instead. */
+  /**
+   * @deprecated Use {@link #getAzureUseMsi()} instead.
+   */
   @Deprecated()
   public boolean getAzureUseMSI() {
     return azureUseMsi;
   }
 
-  /** @deprecated Use {@link #setAzureUseMsi(boolean)} instead. */
+  /**
+   * @deprecated Use {@link #setAzureUseMsi(boolean)} instead.
+   */
   @Deprecated
   public DatabricksConfig setAzureUseMSI(boolean azureUseMsi) {
     this.azureUseMsi = azureUseMsi;
@@ -726,7 +735,7 @@ public class DatabricksConfig {
   }
 
   public DatabricksConfig clone() {
-    return clone(new HashSet<>());
+    return clone(new HashSet<>(Collections.singletonList("logger")));
   }
 
   public DatabricksConfig newWithWorkspaceHost(String host) {
@@ -736,6 +745,7 @@ public class DatabricksConfig {
                 // The config for WorkspaceClient has a different host and Azure Workspace resource
                 // ID, and also omits
                 // the account ID.
+                "logger",
                 "host",
                 "accountId",
                 "azureWorkspaceResourceId",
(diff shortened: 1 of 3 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -755,4 +761,81 @@ public DatabricksConfig newWithWorkspaceHost(String host) {
...
+   * [Internal] Load the Azure tenant ID from the Azure Databricks login page. If the tenant ID is
+   * already set, this method does nothing.
+   */
+  public void loadAzureTenantId() {
+
+    if (!isAzure() || azureTenantId != null || host == null) {
+      return;
+    }
+
+    final String azureAuthEndpoint = "/aad/auth";
```
gopalldb, line 775: We can declare this as constant. Do we need to make this configurable?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -755,4 +761,81 @@ public DatabricksConfig newWithWorkspaceHost(String host) {
...
+  }
+
+  private String getRedirectLocation(String loginUrl) throws IOException {
+
+    Request request = new Request("GET", loginUrl);
+    request.setRedirectionBehavior(false);
+    Response response = getHttpClient().execute(request);
+    int statusCode = response.getStatusCode();
+
+    if (statusCode / 100 != 3) {
```
samikshya-db, line 805: nit : Can we just do a simple check `statusCode!=302`

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/AzureServicePrincipalCredentialsProvider.java
```diff
@@ -22,12 +22,12 @@ public String authType() {
...
     if (!config.isAzure()
         || config.getAzureClientId() == null
-        || config.getAzureClientSecret() == null
-        || config.getAzureTenantId() == null) {
+        || config.getAzureClientSecret() == null) {
       return null;
     }
     AzureUtils.ensureHostPresent(
         config, mapper, AzureServicePrincipalCredentialsProvider::tokenSourceFor);
+    config.loadAzureTenantId();
```
samikshya-db, line 30: can you return null if azure tenant id fails? to be consistent with line 26? 
Also, can we keep the azureSPCredentials specific helpers  for extracting tenant id here in credentialsProvider itself?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -726,7 +731,7 @@ private DatabricksConfig clone(Set<String> fieldsToSkip) {
   }
 
   public DatabricksConfig clone() {
-    return clone(new HashSet<>());
+    return clone(new HashSet<>(Collections.singletonList("logger")));
```
gopalldb, line 734: why is this needed?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -755,4 +761,81 @@ public DatabricksConfig newWithWorkspaceHost(String host) {
   public String getEffectiveOAuthRedirectUrl() {
     return redirectUrl != null ? redirectUrl : "http://localhost:8080/callback";
   }
+
+  /**
+   * [Internal] Load the Azure tenant ID from the Azure Databricks login page. If the tenant ID is
+   * already set, this method does nothing.
+   */
+  public void loadAzureTenantId() {
```
gopalldb, line 769: and if it cannot load, then also it will do nothing? Will the caller expect that tenant-Id is set after this operation?

Instead can we return a boolean on whether the tenant-Id was successfully set?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
sreekanth-db: The `logger` field is excluded from cloning because it's a `static final` field. The `clone(Set<String> fieldsToSkip)` method uses reflection to copy field values with `f.set(newConfig, f.get(this))`. 

Attempting to clone a static field causes an `IllegalAccessException` with the message: "Can not set static final org.slf4j.Logger field com.databricks.sdk.core.DatabricksConfig.logger to org.slf4j.reload4j.Reload4jLoggerAdapter"

The logger is shared across all instances anyway, so excluding it from cloning is the correct approach.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
sreekanth-db: Updated to return `boolean` with clear semantics:
- `true` = tenant ID is available (either was already set or successfully loaded)
- `false` = tenant ID is not available (configuration issues, network failures, etc.)

This provides clear feedback to callers about the operation's success.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
sreekanth-db: Moved to constant `AZURE_AUTH_ENDPOINT`

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -726,7 +735,7 @@ private DatabricksConfig clone(Set<String> fieldsToSkip) {
   }
 
   public DatabricksConfig clone() {
```
nikhilsuri-db, line 737: Do we have to clone this method for all our config classes?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
```diff
@@ -755,4 +766,88 @@ public DatabricksConfig newWithWorkspaceHost(String host) {
...
+   * 
+   * @return true if tenant ID is available (either was already set or successfully loaded), false otherwise
+   */
+  public boolean loadAzureTenantId() {
+
+    if (azureTenantId != null) {
+      return true; // Tenant ID already available - success
+    }
+
+    if (!isAzure() || host == null) {
```
nikhilsuri-db, line 782: Should not the isAzure check be the first check in this method & return false even if `azureTenantId` is having some value?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksConfig.java
sreekanth-db: Could you elaborate on this comment? Are you asking whether we need to implement similar `clone()` methods in other configuration classes, or are you suggesting a different architectural approach for handling object cloning?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/AzureServicePrincipalCredentialsProvider.java
```diff
@@ -22,12 +22,17 @@ public String authType() {
...
+        || config.getAzureClientSecret() == null) {
       return null;
     }
     AzureUtils.ensureHostPresent(
         config, mapper, AzureServicePrincipalCredentialsProvider::tokenSourceFor);
+        
+    boolean tenantIdLoaded = config.loadAzureTenantId();
+    if (!tenantIdLoaded) {
+      return null;
+    }
```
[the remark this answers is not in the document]
sreekanth-db: Updated the code as per your suggestions:
- Not changing the state of `DatabricksConfig`
- Maintaining local variable for `tenantId` in `AzureServicePrincipalCredentialsProvider `
- Moved the tenant id inferring logic to `AzureUtils` for reusability

databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/AzureUtils.java
```diff
@@ -95,4 +103,87 @@ public static Optional<String> getAzureWorkspaceResourceId(Workspace workspace)
...
+      return null;
+    }
+
+    String loginUrl = config.getHost() + AZURE_AUTH_ENDPOINT;
+
+    try {
+      String redirectLocation = getRedirectLocation(config, loginUrl);
+      if (redirectLocation == null) {
+        logger.warn("Failed to get redirect location from Azure auth endpoint: {}", loginUrl);
+        return null;
```
renaudhartert-db, line 129: Let's throw an exception instead, the caller should be the one deciding whether these should be logged or not. Same in other places.

renaudhartert-db (changes requested):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/utils/AzureUtils.java
```diff
@@ -95,4 +103,76 @@ public static Optional<String> getAzureWorkspaceResourceId(Workspace workspace)
...
+    if (!config.isAzure()) {
+      throw new DatabricksException("Cannot infer tenant ID: workspace is not Azure");
+    }
+
+    String loginUrl = config.getHost() + AZURE_AUTH_ENDPOINT;
+
+    try {
+      String redirectLocation = getRedirectLocation(config, loginUrl);
+      String extractedTenantId = extractTenantIdFromUrl(redirectLocation);
+      logger.info("Successfully discovered Azure tenant ID: {}", extractedTenantId);
```
renaudhartert-db, line 133: Also remove `private static final Logger logger = LoggerFactory.getLogger(AzureUtils.class);`

```suggestion
```

renaudhartert-db (approved):
LGTM modulo remaining comment to remove the logger from AzureUtils.

PR #489, opened 2025-07-17
Fix Release Workflow to publish in Maven Central.

parthban-db:

This fixes three things in the release workflow:
- It fixes how we pass the secrets to setup-java. Previously, the secret was passed in the format of {env.Secret}.
- Added a block in the pom file that signs the binary.
- Added the name of the JAVA Library in the pom file.

## How is this tested?

Manually tested.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,10 +28,10 @@ jobs:
           java-version: 8
           server-id: central
           distribution: "adopt"
-          server-username: ${{ secrets.MAVEN_CENTRAL_USERNAME }}
-          server-password: ${{ secrets.MAVEN_CENTRAL_PASSWORD }}
+          server-username: MAVEN_CENTRAL_USERNAME
+          server-password: MAVEN_CENTRAL_PASSWORD
           gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
-          gpg-passphrase: ${{ secrets.GPG_PASSPHRASE }}
+          gpg-passphrase: GPG_PASSPHRASE
 
       # This step runs ONLY on branch pushes (dry-run)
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -28,20 +28,28 @@ jobs:
           java-version: 8
           server-id: central
           distribution: "adopt"
-          server-username: ${{ secrets.MAVEN_CENTRAL_USERNAME }}
-          server-password: ${{ secrets.MAVEN_CENTRAL_PASSWORD }}
+          server-username: MAVEN_CENTRAL_USERNAME
+          server-password: MAVEN_CENTRAL_PASSWORD
           gpg-private-key: ${{ secrets.GPG_PRIVATE_KEY }}
```
hectorcast-db, line 33: What is the issue with the old way?
Why it does not apply to the private key?

.github/workflows/release.yml
```diff
@@ -28,20 +28,28 @@ jobs:
...
-          gpg-passphrase: ${{ secrets.GPG_PASSPHRASE }}
-
+          gpg-passphrase: GPG_PASSPHRASE
+          
       # This step runs ONLY on branch pushes (dry-run)
       - name: Run Release Dry-Run (Verify)
         if: "!startsWith(github.ref, 'refs/tags/')"
         run: mvn -Prelease -DskipTests=true --batch-mode verify
+        env:
```
hectorcast-db, line 40: No private key needed here?

.github/workflows/release.yml
parthban-db: The issue with the current behaviour is that the secret is passed in the format of {env.Secret}. Which corrupts the secret.

PR #496, opened 2025-08-07
Add an option to set a timeout for browser confirmation in the U2M authentication flow. 

renaudhartert-db:

This PR adds a configurable timeout for OAuth browser authentication in the Databricks SDK for Java. 

The OAuth browser authentication flow currently waits indefinitely for user interaction, which can cause applications to hang indefinitely if users don't complete the authentication process.

The key changes include:

- New configuration option: Added `setOAuthBrowserAuthTimeout(Duration timeout)` to `DatabricksConfig` with environment variable support via `DATABRICKS_OAUTH_BROWSER_AUTH_TIMEOUT`.
- Timeout implementation: Modified the OAuth consent flow in Consent.CallbackResponseHandler to enforce the configured timeout when waiting for browser callback parameters.
- Bug fixes: Fixed ConfigAttributeAccessor to properly handle Duration, Integer, and Boolean types from environment variables.

## How is this tested?

Added unit tests to verify that env variables are properly loaded and that the timeout logic works as expected.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Consent.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Consent.java
@@ -388,10 +388,10 @@ public class Consent implements Serializable {
       exchange.close();
     }
 
-    /** 
-     * Wait and return the params. 
-     * 
-     * This method might throw an exception in case of timeout.
+    /**
+     * Wait and return the params.
+     *
+     * <p>This method might throw an exception in case of timeout.
      */
     public Map<String, String> getParams() {
       synchronized (lock) {
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Consent.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Consent.java
@@ -113,10 +113,6 @@ public class Consent implements Serializable {
       return this;
     }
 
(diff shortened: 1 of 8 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Consent.java
```diff
@@ -373,7 +389,16 @@ public Map<String, String> getParams() {
...
           try {
-            lock.wait();
+            if (timeout.isPresent()) {
+              Duration t = timeout.get();
+              lock.wait(t.toMillis());
+              if (params == null) {
+                throw new DatabricksException(
+                    "OAuth browser authentication timed out after " + t.getSeconds() + " seconds");
+              }
+            } else {
```
parthban-db, line 399: I think doing this can leave a connection open that we had started to get the params from the browser. As this will throw an error without waiting for the connection to be closed. 

https://github.com/databricks/databricks-sdk-java/pull/496/files#diff-5149deb8effc0c061128dfe871cd7b8850d479e3a220e3856ff7f4b46e1b6655L226-L229

I think changing the above code will solve the problem, not sure if there are any other cases to consider:
```
    try {
      desktopBrowser();
      return handler.getParams();
    } finally {
      httpServer.stop(0);
    }
```

renaudhartert-db: Good catch

PR #511, opened 2025-09-17
Update SDK to latest API definition

parthban-db:

Update SDK to the latest API definition.

## How is this tested?

Existing CI.


parthban-db:
>I couldn't really understand which manual changes were added for IAM?


Basically, the existing SCIM APIs are added as a handwritten service, as we deleted them from the API definition.

PR #740, opened 2026-03-26
Add logging abstraction with SLF4J backend

mihaimitrea-db:
## 🥞 Stacked PR
Use this [link](https://github.com/databricks/databricks-sdk-java/pull/740/files) to review incremental changes.
- [**stack/logging-abstraction**](https://github.com/databricks/databricks-sdk-java/pull/740) [[Files changed](https://github.com/databricks/databricks-sdk-java/pull/740/files)]
  - [stack/logging-jul](https://github.com/databricks/databricks-sdk-java/pull/741) [[Files changed](https://github.com/databricks/databricks-sdk-java/pull/741/files/5156924f..c04a3808)]
    - [stack/logging-migration](https://github.com/databricks/databricks-sdk-java/pull/742) [[Files changed](https://github.com/databricks/databricks-sdk-java/pull/742/files/c04a3808..d5c03d4f)]

---------
## Summary

Introduces a logging abstraction layer (`com.databricks.sdk.core.logging`) that decouples the SDK's internal logging from any specific backend. SLF4J remains the default. This PR contains only the abstraction and the SLF4J backend; the JUL backend and the call-site migration follow in stacked PRs.

## Why

The SDK currently imports `org.slf4j.Logger` and `org.slf4j.LoggerFactory` directly in every class that logs. This hard coupling means users who embed the SDK in environments where SLF4J is impractical (e.g. BI tools with constrained classpaths) have no way to switch to an alternative logging backend like `java.util.logging`.

We need an indirection layer that lets users swap the logging backend programmatically while keeping SLF4J as the zero-configuration default so that existing users don't have to change anything.

The design follows the pattern established by SLF4J itself and [Netty's `InternalLoggerFactory`](https://netty.io/4.1/api/io/netty/util/internal/logging/InternalLoggerFactory.html): an `ILoggerFactory` interface that backends implement, a `LoggerFactory` utility class with static `getLogger` / `setDefault` methods, and a separate `Logger` abstract class that serves as a clean extension point for custom implementations.

### Interface changes

- **`Logger`** — new abstract class in `com.databricks.sdk.core.logging`. Defines the logging contract: `debug`, `info`, `warn`, `error` (each with plain-string, varargs, and `Supplier<String>` overloads). Users extend this to build custom loggers.
- **`ILoggerFactory`** — new interface. Backends implement `createLogger(Class<?>)` and `createLogger(String)` to produce `Logger` instances. Users implement this to provide a fully custom logging backend.
- **`LoggerFactory`** — new `final` utility class. Static methods `getLogger(Class<?>)` and `getLogger(String)` return loggers from the current default factory. `setDefault(ILoggerFactory)` overrides the backend — must be called before creating any SDK client.
- **`Slf4jLoggerFactory`** — public concrete `ILoggerFactory` implementation with a singleton `INSTANCE`. This is the default.

### Behavioral changes

None. SLF4J is the default backend and all logging calls pass through to `org.slf4j.Logger` exactly as before. Existing users see no difference.

### Internal changes

- **`Slf4jLogger`** — package-private class that delegates all calls to an `org.slf4j.Logger`. Fully qualifies `org.slf4j.LoggerFactory` references to avoid collision with the SDK's `LoggerFactory`.
- All new classes live in `com.databricks.sdk.core.logging`.

## How is this tested?

- `LoggerFactoryTest` — verifies the default factory is SLF4J, that `setDefault(null)` is rejected, and that `getLogger(String)` works.
- `Slf4jLoggerTest` — verifies `LoggerFactory.getLogger` returns the correct type, exercises all logging methods including varargs and trailing Throwable via a capturing Log4j appender that asserts on message content, level, and attached throwable.
- Full test suite passes.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/logging/LoggerFactory.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/logging/LoggerFactory.java
@@ -1,0 +1,65 @@
+package com.databricks.sdk.core.logging;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Creates and configures {@link Logger} instances for the SDK.
+ *
+ * <p>By default, logging goes through SLF4J. Users can override the backend programmatically before
+ * creating any SDK client:
+ *
+ * <pre>{@code
+ * LoggerFactory.setDefault(myCustomFactory);
+ * WorkspaceClient ws = new WorkspaceClient();
+ * }</pre>
+ *
+ * <p>Extend this class to provide a fully custom logging backend.
+ */
+public abstract class LoggerFactory {
+
+  private static final AtomicReference<LoggerFactory> defaultFactory = new AtomicReference<>();
+
+  /** Returns a logger for the given class, using the current default factory. */
+  public static Logger getLogger(Class<?> type) {
+    return getDefault().createLogger(type);
+  }
+
+  /** Returns a logger with the given name, using the current default factory. */
+  public static Logger getLogger(String name) {
+    return getDefault().createLogger(name);
+  }
+
+  /**
+   * Overrides the logging backend used by the SDK.
+   *
+   * <p>Must be called before creating any SDK client or calling {@link #getLogger}. Loggers already
+   * obtained will not be affected by subsequent calls.
+   */
+  public static void setDefault(LoggerFactory factory) {
+    if (factory == null) {
+      throw new IllegalArgumentException("LoggerFactory must not be null");
+    }
+    defaultFactory.set(factory);
+  }
+
+  static LoggerFactory getDefault() {
+    LoggerFactory f = defaultFactory.get();
+    if (f != null) {
+      return f;
+    }
+    defaultFactory.compareAndSet(null, Slf4jLoggerFactory.INSTANCE);
+    return defaultFactory.get();
+  }
+
+  /**
+   * Creates a {@link Logger} for the given class. Subclasses obtain the backend logger (e.g. SLF4J)
+   * and return an adapter.
+   */
+  protected abstract Logger createLogger(Class<?> type);
+
+  /**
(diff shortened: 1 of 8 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/logging/LoggerFactory.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * <p>Extend this class to provide a fully custom logging backend.
+ */
+public abstract class LoggerFactory {
+
+  private static final AtomicReference<LoggerFactory> defaultFactory = new AtomicReference<>();
+
+  /** Returns a logger for the given class, using the current default factory. */
+  public static Logger getLogger(Class<?> type) {
+    return getDefault().newInstance(type);
```
parthban-db: Does newInstance always create a new logger? What will happen when a logger exists for a given class?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/logging/Slf4jLogger.java
```diff
@@ -0,0 +1,85 @@
...
+  @Override
+  public void info(String format, Object... args) {
+    delegate.info(format, args);
+  }
+
+  @Override
+  public void info(Supplier<String> msgSupplier) {
+    if (delegate.isInfoEnabled()) {
+      delegate.info(msgSupplier.get());
+    }
```
parthban-db, line 45: Do we need delegate.isInfoEnabled check everywhere?

parthban-db (approved):

databricks-sdk-java/src/main/java/com/databricks/sdk/core/logging/Slf4jLogger.java
mihaimitrea-db: The supplier function is reserved for logs which are heavy on the computation side so you would like to avoid computing them if the logger is not configured for that.

PR #751, opened 2026-03-31
Fix CLI token source --profile fallback with version detection

mihaimitrea-db:
## 🥞 Stacked PR
Use this [link](https://github.com/databricks/databricks-sdk-java/pull/751/files) to review incremental changes.
- [**stack/cli-force-refresh**](https://github.com/databricks/databricks-sdk-java/pull/751) [[Files changed](https://github.com/databricks/databricks-sdk-java/pull/751/files)]
  - [stack/cli-attempt-chain](https://github.com/databricks/databricks-sdk-java/pull/752) [[Files changed](https://github.com/databricks/databricks-sdk-java/pull/752/files/c4e12c18..5ee21871)]

---------
## Summary

Replace the broken error-based `--profile` fallback in `CliTokenSource` with version-based CLI detection at init time. Mirrors [databricks/databricks-sdk-go#1605](https://github.com/databricks/databricks-sdk-go/pull/1605) and [databricks/databricks-sdk-py#1377](https://github.com/databricks/databricks-sdk-py/pull/1377).

## Why

`--profile` on `databricks auth token` is a global flag, so old CLIs (< v0.207.1) **silently accept it** and then fail with `"cannot fetch credentials"` instead of `"unknown flag: --profile"`. The existing retry check was matching on the latter and never fired — the `--host` fallback it gated was effectively dead code. Switching to `databricks version` + a minimum-version constant makes the fallback reliable and sets up future capability-gated flags (e.g. `--force-refresh` in [#752](https://github.com/databricks/databricks-sdk-java/pull/752)) without additional subprocess calls.

### Interface changes

None. `CliTokenSource` is not part of the public API surface.

### Behavioral changes

- `cfg.profile` + CLI < v0.207.1 now correctly falls back to `--host` (previously broken).
- `databricks version` failures log a `WARNING` and fall back to the most conservative command. Successful detections are cached per CLI path; failures are not cached and will be retried on the next call.
- A default dev build (`v0.0.0-dev`) logs an `INFO` explaining why feature gates are conservative.

`AzureCliCredentialsProvider` is untouched.

### Internal changes

- New `DatabricksCliVersion` class with a `(major, minor, patch)` triple, an `UNKNOWN` sentinel, an `atLeast()` comparator, and an `isDefaultDevBuild()` helper.
- `CliTokenSource` simplified to a single `cmd`; the `fallbackCmd` parameter and its retry logic are removed.
- `DatabricksCliCredentialsProvider` gains `getCliVersion`, `probeCliVersion`, `parseCliVersion`, `resolveCliCommand`, and `buildCliCommand` helpers.

## How is this tested?

Unit tests in `DatabricksCliVersionTest` cover version comparison (across patch/minor/major), the `UNKNOWN` sentinel, dev-build detection, and `toString` formatting.

Unit tests in `DatabricksCliCredentialsProviderTest` cover JSON parsing of `databricks version --output json` (standard, dev build, missing fields, malformed JSON, empty string) and command assembly for every profile/host/version combination (host-only, account host, profile + new CLI, profile + old CLI, unknown version, dev build).

`CliTokenSourceTest` retains its parsing and timezone tests; the obsolete fallback tests are dropped.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
@@ -6,12 +6,20 @@ import com.databricks.sdk.core.oauth.CachedTokenSource;
 import com.databricks.sdk.core.oauth.OAuthHeaderFactory;
 import com.databricks.sdk.core.oauth.Token;
 import com.databricks.sdk.core.oauth.TokenSource;
+import com.databricks.sdk.core.utils.Environment;
 import com.databricks.sdk.core.utils.OSUtils;
 import com.databricks.sdk.support.InternalApi;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Paths;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import org.apache.commons.io.IOUtils;
 
 @InternalApi
 public class DatabricksCliCredentialsProvider implements CredentialsProvider {
@@ -21,6 +29,28 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
   public static final String DATABRICKS_CLI = "databricks-cli";
 
   private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  // ---- Version detection ----
+
+  // --profile support added in CLI v0.207.1: https://github.com/databricks/cli/pull/855
+  static final DatabricksCliVersion CLI_VERSION_FOR_PROFILE = new DatabricksCliVersion(0, 207, 1);
+
+  // 5-second cap on `databricks version` so a hung CLI (slow first-run scan, antivirus, blocked
+  // stdin) does not wedge SDK init indefinitely.
+  private static final long VERSION_PROBE_TIMEOUT_SECONDS = 5;
+
+  // Successful version probes keyed by cliPath. Subprocess failures (timeouts, non-zero exit,
+  // IO errors) and probes that returned UNKNOWN due to unparseable output are deliberately not
+  // cached, so a transient error does not pin every later token source to the conservative
+  // fallback for the rest of the process lifetime.
+  private static final Map<String, DatabricksCliVersion> VERSION_CACHE = new ConcurrentHashMap<>();
+
+  /** Test-only hook to clear the cross-test version cache. Package-private. */
+  static void clearVersionCache() {
+    VERSION_CACHE.clear();
+  }
+
+  // ---- Scope validation ----
 
   /** Thrown when the cached CLI token's scopes don't match the SDK's configured scopes. */
   static class ScopeMismatchException extends DatabricksException {
@@ -36,57 +66,11 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
   private static final Set<String> SCOPES_IGNORED_FOR_COMPARISON =
       Collections.singleton("offline_access");
 
+  // ---- Public API ----
+
   @Override
   public String authType() {
     return DATABRICKS_CLI;
-  }
-
-  /**
-   * Builds the CLI command arguments using --host (legacy path).
-   *
-   * @param cliPath Path to the databricks CLI executable
-   * @param config Configuration containing host, account ID, workspace ID, etc.
-   * @return List of command arguments
-   */
-  List<String> buildHostArgs(String cliPath, DatabricksConfig config) {
-    List<String> cmd =
-        new ArrayList<>(Arrays.asList(cliPath, "auth", "token", "--host", config.getHost()));
-    if (config.getClientType() == ClientType.ACCOUNT) {
-      cmd.add("--account-id");
-      cmd.add(config.getAccountId());
-    }
-    return cmd;
-  }
-
-  private CliTokenSource getDatabricksCliTokenSource(DatabricksConfig config) {
-    String cliPath = config.getDatabricksCliPath();
-    if (cliPath == null) {
-      cliPath = OSUtils.get(config.getEnv()).getDatabricksCliPath();
-    }
-    if (cliPath == null) {
-      LOG.debug("Databricks CLI could not be found");
-      return null;
-    }
-
-    List<String> cmd;
-    List<String> fallbackCmd = null;
-
-    if (config.getProfile() != null) {
-      // When profile is set, use --profile as the primary command.
-      // The profile contains the full config (host, account_id, etc.).
-      cmd =
-          new ArrayList<>(
-              Arrays.asList(cliPath, "auth", "token", "--profile", config.getProfile()));
-      // Build a --host fallback for older CLIs that don't support --profile.
-      if (config.getHost() != null) {
-        fallbackCmd = buildHostArgs(cliPath, config);
-      }
-    } else {
-      cmd = buildHostArgs(cliPath, config);
-    }
-
-    return new CliTokenSource(
-        cmd, "token_type", "access_token", "expiry", config.getEnv(), fallbackCmd);
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


simonfaltum (commented):
## Review Swarm (Isaac + Cursor, 2 rounds, deep review)

### Verdict: **Not ready yet**

**0** Critical | **2** Major | **1** Gap (Minor) | **8** Nit | **6** Suggestion

Both reviewers independently proposed the same architecture the PR already uses (probe + cache + typed semver + UNKNOWN sentinel). **The two Major findings are about subprocess and JSON-parsing robustness, not the design.** Everything else is polish.

See inline comments for specific findings, and a follow-up general PR comment for findings that reference lines outside the diff hunks.

### Verifications (both reviewers agreed, no issues)
- `CliTokenSource` carries `@InternalApi`; removing the 6-arg constructor is not a breaking change.
- `AzureCliCredentialsProvider` is untouched (still uses the 5-arg constructor).
- `atLeast`/`compareTo` ordering is correct across patch/minor/major; `UNKNOWN(-1,-1,-1)` sorts below all real versions including `(0,0,0)`.
- Account-host vs workspace-host behavior is preserved (`buildHostArgs` still appends `--account-id` only when `ClientType.ACCOUNT`).
- **The core bug is fixed**: `cfg.profile` set with CLI < v0.207.1 now correctly falls back to `--host`.

---

_Automated review swarm. Classifications are advisory; final judgement is yours._

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +128,187 @@ public Token getToken() {
...
+      throw new IOException("interrupted waiting for `databricks version`", e);
+    }
+
+    if (process.exitValue() != 0) {
+      String stderr = readStream(process.getErrorStream());
+      throw new IOException(
+          "`databricks version` exited with code " + process.exitValue() + ": " + stderr);
+    }
+
+    return parseCliVersion(readStream(process.getInputStream()));
```
simonfaltum: **[Major]** `probeCliVersion` reads streams *after* `waitFor` — fragile pipe-buffer pattern

**Found by:** Both (Isaac: Major; Cursor: Suggestion→Minor)

`probeCliVersion` calls `process.waitFor(5s)` *before* draining stdout/stderr via `readStream`. If the child ever writes more than the OS pipe buffer (~64 KB on Linux/macOS, smaller on Windows) on either stream, the child blocks writing and `waitFor` necessarily times out. Three aggravating factors:

1. The sibling `CliTokenSource.execCliCommand` already drains streams *before* waiting — this PR introduces a less-safe pattern in adjacent code.
2. The pattern is asymmetric: on success only stdout is read; on non-zero exit only stderr is read. The undrained stream is left attached to the (now-exited) process until GC.
3. This probe runs on every init and future capability flags (e.g. `--force-refresh` in stacked #752) depend on it.

Today's `databricks version --output json` payload is ~150 bytes so this is latent, not a live bug — which is why the reviewers disagree on severity. Kept at Major because fixing it now also resolves three related Nits (stream close, stdin leak, InterruptedException symmetry) in one shot.

**Suggestion:** Use `pb.redirectErrorStream(true)` plus a single async stream drain, then `waitFor(timeout, SECONDS)`. Close streams explicitly in a `finally` block. Also close `process.getOutputStream()` (stdin).

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +128,187 @@ public Token getToken() {
...
+      JsonNode node = MAPPER.readTree(output);
+      JsonNode major = node.get("Major");
+      JsonNode minor = node.get("Minor");
+      JsonNode patch = node.get("Patch");
+      if (major == null || minor == null || patch == null) {
+        LOG.debug(
+            "Failed to parse Databricks CLI version: missing Major/Minor/Patch in {}", output);
+        return DatabricksCliVersion.UNKNOWN;
+      }
+      return new DatabricksCliVersion(major.asInt(), minor.asInt(), patch.asInt());
```
simonfaltum, line 362: **[Major]** `parseCliVersion` silently coerces non-integer Major/Minor/Patch to `(0,0,0)` — collides with dev-build sentinel

**Found by:** Isaac, confirmed by Cursor (both Major)

`JsonNode.asInt()` returns `0` with no exception for strings, booleans, explicit JSON `null`, objects, and arrays. So `{"Major":"abc","Minor":"x","Patch":"y"}` parses to `(0,0,0)`, which the rest of the code interprets as the **default dev build** — a valid CLI that lacks version metadata. The user gets an INFO "development build" message and conservative fallbacks, with no hint that the JSON payload was actually garbage. Explicit JSON `null` for any field (`NullNode`) also silently becomes `0`.

The existing `testParseCliVersion_MalformedJson` covers whole-string failure but not field-level type confusion. Real CLI output won't hit this today (Go's side serializes ints), but the parser's whole job is to be defensive against format drift.

**Suggestion:**

```java
if (!major.isIntegralNumber() || !minor.isIntegralNumber() || !patch.isIntegralNumber()) {
  return DatabricksCliVersion.UNKNOWN;
}
```

Add parameterized test cases for: string values, explicit JSON `null`, arrays/objects, and a mix of valid + invalid fields.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +128,187 @@ public Token getToken() {
...
+    }
+
+    try {
+      DatabricksCliVersion version = probeCliVersion(cliPath, env);
+      VERSION_CACHE.put(cliPath, version);
+      return version;
+    } catch (Exception e) {
+      LOG.warn(
+          "Failed to detect Databricks CLI version: {}. Falling back to conservative flag set.",
+          e.getMessage());
```
simonfaltum: **[Nit]** SLF4J `LOG.warn` drops the exception stack trace

**Found by:** Cursor, confirmed by Isaac

SLF4J only logs the throwable when the *trailing* vararg is itself a `Throwable`. As written, `LOG.warn("...: {}...", e.getMessage())` only formats the message into `{}` — the stack trace is lost entirely. Debugging real-world probe failures (timeouts, AV interference, exec permission issues) becomes much harder.

**Suggestion:** Add `e` as a trailing arg:

```java
LOG.warn("Failed to detect Databricks CLI version: {}. Falling back to conservative flag set.",
    e.getMessage(), e);
```

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +128,187 @@ public Token getToken() {
...
+    DatabricksCliVersion cached = VERSION_CACHE.get(cliPath);
+    if (cached != null) {
+      return cached;
+    }
+
+    try {
+      DatabricksCliVersion version = probeCliVersion(cliPath, env);
+      VERSION_CACHE.put(cliPath, version);
+      return version;
+    } catch (Exception e) {
```
simonfaltum: **[Nit]** `catch (Exception e)` swallows RuntimeExceptions, hiding programming errors

**Found by:** Isaac, confirmed by Cursor

`probeCliVersion` only declares `throws IOException`, so catching `Exception` also swallows `RuntimeException` (NPE, ISE, `SecurityException` from `pb.start()`). Unrelated bugs get silently converted into "conservative fallback" behavior — exactly the debugging nightmare the PR is supposed to end.

**Suggestion:** `catch (IOException e)` and let RuntimeExceptions surface. If you want defensive coverage of `pb.start()`'s unchecked throws, list them explicitly: `catch (IOException | SecurityException e)`.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -22,6 +29,22 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
...
+  static final DatabricksCliVersion CLI_VERSION_FOR_PROFILE = new DatabricksCliVersion(0, 207, 1);
+
+  // 5-second cap on `databricks version` so a hung CLI (slow first-run scan, antivirus, blocked
+  // stdin) does not wedge SDK init indefinitely.
+  private static final long VERSION_PROBE_TIMEOUT_SECONDS = 5;
+
+  // Successful version probes keyed by cliPath. Failures are deliberately not cached, so a
+  // transient error (timeout, AV scan) does not pin every later token source to the conservative
+  // fallback for the rest of the process lifetime.
+  private static final Map<String, DatabricksCliVersion> VERSION_CACHE = new ConcurrentHashMap<>();
```
simonfaltum, line 46: **[Nit]** `VERSION_CACHE` has no test reset hook

**Found by:** Both

This cache is process-global with no way to clear it. Today's tests only exercise pure functions (`parseCliVersion`, `buildCliCommand`) so they don't hit it, but as soon as anyone adds unit tests for `getCliVersion` or writes an integration test crossing two configs with the same `cliPath`, state leaks across test classes. JUnit 5 reorders class execution — this becomes flaky.

**Suggestion:** Add a package-private `@VisibleForTesting static void clearVersionCache()` and call it in new tests' `@BeforeEach`.

simonfaltum: **[Nit]** "Failures are not cached" slightly overclaims cache recovery

**Found by:** Isaac, confirmed by Cursor

Exceptions out of `probeCliVersion` aren't cached — true. But a *successful* probe that returns `UNKNOWN` (parseable-but-malformed JSON, field-level type confusion from the `parseCliVersion` Major above) **is** cached. So "transient subprocess timeout" recovers on the next call, but "CLI returns ambiguous JSON" does not.

**Suggestion:** Reword to make the distinction explicit:

> Subprocess failures (timeouts, non-zero exit, IO errors) are not cached. Successful probes, including those that returned UNKNOWN due to unparseable output, are cached.

Same clarification in the PR description.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +128,187 @@ public Token getToken() {
...
+        process.destroyForcibly();
+        throw new IOException(
+            "timed out after "
+                + VERSION_PROBE_TIMEOUT_SECONDS
+                + "s waiting for `databricks version`");
+      }
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new IOException("interrupted waiting for `databricks version`", e);
+    }
```
simonfaltum, line 298: **[Nit]** `InterruptedException` path leaks the CLI subprocess

**Found by:** Isaac, confirmed by Cursor

When `waitFor(timeout)` is interrupted, the code restores the interrupt flag and throws `IOException`, but never destroys the subprocess. The `databricks version` process keeps running until it exits on its own. Compare to the timeout path (line 256) which correctly calls `process.destroyForcibly()` — the patterns should be symmetric.

**Suggestion:** Call `process.destroyForcibly()` inside the `catch (InterruptedException e)` block before rethrowing.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
simonfaltum: **[Nit]** `probeCliVersion` leaks process streams (stdout, stderr, stdin)

**Found by:** Both

`IOUtils.toByteArray(stream)` in `readStream` (line 306-307) does not close the stream. The timeout path never reads them at all. Process stdin (`process.getOutputStream()`) is also never closed — `databricks version` doesn't read stdin so no hang in practice, but the FD is leaked until `Process` is GC'd.

This stacks with the Major stream-drain finding: using `pb.redirectErrorStream(true)` plus explicit `close()` in `finally` resolves all three leaks together.

**Suggestion:** Wrap stream reads in try-with-resources, close in `finally` alongside `destroyForcibly()` for the timeout path. Optionally use `pb.redirectInput(Redirect.DISCARD)` to close stdin explicitly at process start.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +128,187 @@ public Token getToken() {
...
+  }
+
+  // ---- Version detection ----
+
+  /**
+   * Returns the CLI version, catching subprocess failures so the caller can proceed with the
+   * conservative fallback. Successful results are cached per {@code cliPath} for the process
+   * lifetime; failures are not cached and will be retried on the next call.
+   */
+  DatabricksCliVersion getCliVersion(String cliPath, Environment env) {
```
simonfaltum, line 240: **[Gap (Minor)]** No unit tests for `getCliVersion` cache behavior, `probeCliVersion` error paths, or `resolveCliCommand` glue

**Found by:** Both

Tests cover `parseCliVersion` and `buildCliCommand` (pure functions — easy) but skip the methods most likely to misbehave: `getCliVersion` (cache + exception swallowing), `probeCliVersion` (subprocess), and `resolveCliCommand` (integration glue). The PR description specifically advertises caching semantics — "successful detections are cached per CLI path; failures are not cached and will be retried" — but no test verifies this.

`probeCliVersion` is package-private and overridable, so tests can subclass with a fake. No `MockedConstruction<ProcessBuilder>` is needed.

**Suggestion:** Add tests for:
1. Successful probe caches — second call returns cached value without re-probing.
2. Thrown probe returns UNKNOWN and does *not* cache — second call re-probes.
3. Probe with non-zero exit returns UNKNOWN.
4. `resolveCliCommand` with no profile does not trigger a probe (ties to the host-only fast-path Suggestion below).
5. The full `resolveCliCommand` matrix: {new CLI, old CLI, dev build, UNKNOWN} × {profile, no profile}.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +128,187 @@ public Token getToken() {
...
+    DatabricksCliVersion version = getCliVersion(cliPath, config.getEnv());
+    if (version.isDefaultDevBuild()) {
+      // A default-marker dev build has no injected version, so every feature gate fails.
+      // Surface an informational hint so users know why their feature flags aren't taking effect.
+      LOG.info(
+          "Databricks CLI {} is a development build; feature detection will use conservative "
+              + "fallbacks. Rebuild the CLI with an explicit version to enable capability-based "
+              + "flag selection.",
+          version);
+    }
```
simonfaltum: **[Nit]** Dev-build INFO log fires even on the host-only path where no gate matters

**Found by:** Both

This INFO fires whenever version is `(0,0,0)`, regardless of whether `cfg.profile` is set. For host-only auth no feature gate is consulted, so "conservative fallbacks" is misleading — nothing was actually downgraded.

Related asymmetry: UNKNOWN gets a WARN from `getCliVersion` and a WARN from `buildCliCommand` only when `cfg.profile` is set. Users with `cfg.profile=null` and a probe failure see no explanation of why gates are off.

**Suggestion:** Move the dev-build INFO into `buildCliCommand` next to the existing `version.equals(UNKNOWN) || version.isDefaultDevBuild()` branch so it only fires when a gate is actually skipped. Or accept the noise once #752 lands with `--force-refresh`. Consider mirroring the INFO for UNKNOWN in the same spot for symmetry.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +128,187 @@ public Token getToken() {
...
+    return buildCliCommand(cliPath, config, version);
+  }
```
simonfaltum, line 161: **[Suggestion]** Host-only CLI auth runs an unnecessary version probe

**Found by:** Cursor (cross-review)

`resolveCliCommand` unconditionally calls `getCliVersion(cliPath, config.getEnv())`, but `buildCliCommand` immediately returns `buildHostArgs(...)` when `config.getProfile() == null`. Host-only CLI auth has no version-gated flag in this PR, so every host-only token-source initialization pays for a `databricks version --output json` subprocess it doesn't need. If the probe is slow, host-only auth waits up to 5s before doing work it never needed; if the probe fails, the SDK logs a warning even though `databricks auth token --host ...` would have succeeded.

Stacked PR #752 may make the probe universally required via `--force-refresh` — worth confirming.

**Suggestion:** Fast-path host-only command construction before probing: return `buildHostArgs(cliPath, config)` immediately when `config.getProfile() == null` (and no other version-gated options are requested). Add a test asserting no probe runs in the host-only path.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +128,187 @@ public Token getToken() {
...
+      DatabricksCliVersion version = probeCliVersion(cliPath, env);
+      VERSION_CACHE.put(cliPath, version);
+      return version;
+    } catch (Exception e) {
+      LOG.warn(
+          "Failed to detect Databricks CLI version: {}. Falling back to conservative flag set.",
+          e.getMessage());
+      return DatabricksCliVersion.UNKNOWN;
+    }
+  }
```
simonfaltum, line 269: **[Suggestion]** Race window allows duplicate `databricks version` probes

**Found by:** Both

The `get` → `probe` → `put` sequence is not atomic. Two threads calling `getCliVersion` concurrently for the same `cliPath` will both miss the cache, both spawn `databricks version`, and both `put` the same value. Result is correct but each subprocess costs ~tens of ms (more if AV scans on launch). Bundles/tools spinning up many `WorkspaceClient`s in parallel at startup see N probes instead of 1.

**Suggestion:** Use `computeIfAbsent` with a wrapper that converts `IOException` to `UNKNOWN`:

```java
return VERSION_CACHE.computeIfAbsent(cliPath, p -> probeOrUnknown(p, env));
```

Note: `ConcurrentHashMap.computeIfAbsent` holds a per-bucket lock during the lambda, so if the 5s probe blocks, other keys still proceed but another caller for the same key waits. That's usually desirable here (single-flight). If you prefer the current semantics, add a one-line comment that duplicate probes are deliberately acceptable.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
simonfaltum, line 46: **[Suggestion]** Negative-result caching trade-off could bite SDK consumers that build many `WorkspaceClient`s

**Found by:** Cursor, confirmed by Isaac

Failing-fast is fine when the CLI is truly unavailable (fast exception). But a CLI that genuinely *hangs* (corporate AV scan on first launch, blocked exec) incurs **5s on every `configure(...)` call** for the JVM lifetime. For SDK consumers that build many `WorkspaceClient`s this is death by a thousand cuts. Worth confirming parity with the Go/Python siblings.

**Suggestion:** Either (a) document the trade-off explicitly in the cache comment ("we accept up to N×5s startup latency under sustained CLI failure"), or (b) cache UNKNOWN with a small TTL (e.g. 60s) to bound the damage. Not a blocker.

simonfaltum, line 46: **[Suggestion]** Cache key is raw `cliPath` — Windows path-case can split entries

**Found by:** Cursor, confirmed by Isaac

On Windows, the same binary may be referenced as `C:\Program Files\Databricks\databricks.exe` vs `c:\program files\databricks\databricks.exe` (env vs config), giving two cache entries and two probes. Trivial cost in practice; worth a one-line normalization.

**Suggestion:** Normalize via `Paths.get(cliPath).toAbsolutePath().normalize().toString()`, or document the assumption. Skip if you don't want to touch it.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliVersion.java
```diff
@@ -0,0 +1,93 @@
...
+ *       version metadata. See {@link #isDefaultDevBuild()}.
+ *   <li>anything else — a real CLI version.
+ * </ul>
+ *
+ * <p>Prerelease tags are deliberately ignored: feature gates are release-based, so a prerelease of
+ * a version with a flag is assumed to have the flag too.
+ */
+@InternalApi
+public final class DatabricksCliVersion implements Comparable<DatabricksCliVersion> {
+  public static final DatabricksCliVersion UNKNOWN = new DatabricksCliVersion(-1, -1, -1);
```
simonfaltum, line 24: **[Nit]** No named constant for the dev-build sentinel

**Found by:** Cursor, confirmed by Isaac

`UNKNOWN` is a named sentinel; `(0, 0, 0)` is the dev-build sentinel but is referenced by ad-hoc construction (in `parseCliVersion`, `isDefaultDevBuild`, and tests). Symmetry would improve readability and reduce the chance that a refactor silently changes the sentinel value.

**Suggestion:** Add alongside `UNKNOWN`:

```java
public static final DatabricksCliVersion DEFAULT_DEV_BUILD = new DatabricksCliVersion(0, 0, 0);
```

Use it in `parseCliVersion`, `isDefaultDevBuild` (`equals(DEFAULT_DEV_BUILD)`), and in tests. Cosmetic only.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliVersion.java
```diff
@@ -0,0 +1,93 @@
...
+
+  private final int major;
+  private final int minor;
+  private final int patch;
+
+  public DatabricksCliVersion(int major, int minor, int patch) {
```
simonfaltum: **[Nit]** `DatabricksCliVersion.UNKNOWN` is not strictly a singleton

**Found by:** Isaac (cross-review)

The constructor is public, so `new DatabricksCliVersion(-1, -1, -1)` creates a "fake UNKNOWN" that is `.equals(UNKNOWN)` but not `==`. `toString()` and tests use `equals` correctly, but a future consumer doing `version == UNKNOWN` would silently fail. Low-risk because the class is `@InternalApi`.

**Suggestion:** Make the constructor package-private, or reject `(-1, -1, -1)` from public construction. Not urgent.

simonfaltum:
## Review Swarm: Additional findings

The findings below reference code locations outside the PR's diff hunks (pre-existing lines that weren't touched, or test-coverage gaps that don't have a single pin-point) so they're posted here rather than as inline comments.

---

### [Nit] Dead `CliCommandException` inner class + misleading comment after fallback removal
**File:** `databricks-sdk-java/src/main/java/com/databricks/sdk/core/CliTokenSource.java` (class at lines 37-48, comment at lines 121-122)
**Found by:** Cursor, confirmed by Isaac

The PR removes the only consumer of `CliCommandException.getFullOutput()` (the `"unknown flag: --profile"` text check in `getToken`). The class is still thrown from `execCliCommand`, but `fullOutput` is never read — it's effectively a renamed `IOException` with a dead field. Worse, the surrounding comment ("getFullOutput() exposes both streams so the caller can check for `unknown flag: --profile`") is now actively misleading about why the class exists.

**Suggestion:** Either inline to `throw new IOException("cannot get access token: " + stderr)` or, if the dual-stream message is still useful for diagnostics, drop the `getFullOutput()` accessor and update the comment. Cleanup atomic with the obsolescence keeps the diff self-contained.

---

### [Nit] "CLI not installed" path now emits two `WARN` logs instead of one
**File:** `databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java` (existing ~lines 117-122 + new line 235-237)
**Found by:** Isaac (cross-review)

When the CLI binary doesn't exist at `cliPath`, `probeCliVersion` fails first (`ProcessBuilder.start()` throws "error=2, No such file or directory"), `getCliVersion` swallows and emits WARN "Failed to detect Databricks CLI version...". Then `configure` proceeds, `cachedTokenSource.getToken()` runs the same nonexistent binary again, and the existing `configure` catch block emits WARN "Most likely databricks CLI is not installed". **Pre-PR: one WARN. Post-PR: two WARN lines for the same condition.**

**Suggestion:** Downgrade the probe-failure log to DEBUG, or check `IOException.getMessage()` for `"No such file"`/`"error=2"` before warning, so the existing "not installed" WARN remains the canonical signal.

---

### [Nit] `Comparable.compareTo` is not directly exercised by tests
**File:** `databricks-sdk-java/src/test/java/com/databricks/sdk/core/DatabricksCliVersionTest.java`
**Found by:** Cursor, confirmed by Isaac

`atLeast` is tested thoroughly; `compareTo` is only covered transitively. A future refactor that splits the two could regress silently. No test asserts the correct *sign* of `compareTo` (only the boolean result via `atLeast`).

**Suggestion:** Add one parameterized test:

```java
assertEquals(expectedSign, Integer.signum(a.compareTo(b)));
```

covering: equal, lesser-by-major, lesser-by-minor, lesser-by-patch, UNKNOWN-vs-real, dev-vs-real.

---

### [Nit] Wildcard `java.util.*` mixed with explicit `java.util.concurrent.*` imports
**File:** `databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java` (line 18)
**Found by:** Isaac (cross-review)

Pre-existing `import java.util.*;` was kept; new imports (`ConcurrentHashMap`, `TimeUnit`) were added explicitly. Minor inconsistency — the wildcard predates this PR. Defer to repo lint settings; skip if wildcard imports are allowed.

---

### [Suggestion] `buildCliCommand` silently emits `--host null` when host isn't set
**File:** `databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java:171-198`
**Found by:** Cursor; Isaac downgraded to Nit in cross-review

`buildCliCommand` no longer guards `config.getHost() != null` before emitting `--host`. In practice the path is unreachable: `configure(...)` short-circuits with `return null` when `host == null` *before* `getDatabricksCliTokenSource` is called, and `getDatabricksCliTokenSource` is private. Defensive-only. The same latent bug existed pre-PR in the `else` branch.

**Suggestion:** Optional — guard with `Objects.requireNonNull(config.getHost(), ...)` in `buildCliCommand`, or throw a clear `DatabricksException`, so the code documents its own assumption. Keep as-is if relying on `configure`'s short-circuit is fine.

simonfaltum:
Process resource leak in `probeCliVersion` on InterruptedException path
- Priority: P2
- Location: DatabricksCliCredentialsProvider.java: `InterruptedException` catch block inside `probeCliVersion`, after `process.waitFor(VERSION_PROBE_TIMEOUT_SECONDS, TimeUnit.SECONDS)`
- Scenario: Thread is interrupted while waiting for the `databricks version` subprocess. The catch block re-interrupts the thread and throws `IOException` but never calls `process.destroy()` or `process.destroyForcibly()`, leaving the subprocess running. The timeout branch in the same method already calls `process.destroyForcibly()` correctly.
- Potential solution: Wrap the post-`pb.start()` block in a `try/finally` that calls `process.destroyForcibly()` (guarded by `process.isAlive()`), or at minimum add `process.destroyForcibly()` to the `InterruptedException` catch before rethrowing, mirroring the existing timeout branch.

Duplicate WARN log emitted when CLI binary is missing
- Priority: P3
- Location: DatabricksCliCredentialsProvider.java: `getCliVersion` catch block (WARN "Failed to detect Databricks CLI version…") and `configure` catch block (WARN "Most likely databricks CLI is not installed")
- Scenario: `cliPath` is non-null but the binary does not exist. `probeCliVersion` throws (e.g. "No such file or directory"); `getCliVersion` catches it and logs WARN before returning `UNKNOWN`. `getDatabricksCliTokenSource` still constructs a `CliTokenSource`; `configure` calls `getToken()`, the same missing binary fails again, and the existing handler logs the second WARN when the message contains "not found". Pre-PR only one WARN was emitted for this condition.
- Potential solution: Downgrade the version-probe failure log in `getCliVersion` to DEBUG, or detect missing-binary errors specifically and suppress the WARN there, letting the existing "not installed" WARN in `configure` remain the canonical signal.

---
🔍 Reviewed by [nitpicker](https://github.com/arsenyinfo/nitpicker)

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
mihaimitrea-db: Should we really handle cases where the OS pipe buffer overflows? The output of the `cli version` command should be extremely small. Not even close to the upper limit.

simonfaltum: I agree that this is probably over-classified, and the agents has exaggerated the severity

simonfaltum (approved):
LGTM. Thanks for the thorough iteration on the inline feedback — version-gated detection, single-flight caching, parse hardening, and subprocess cleanup all look good now.

One optional nit, take it or leave it: `resolveCliCommand` still runs `getCliVersion` unconditionally, but `buildCliCommand` short-circuits on `profile == null` and ignores the version. Host-only configs pay for a `databricks version` subprocess they never use. Easy fix is to short-circuit before probing. Not blocking, happy to approve as-is.

PR #752, opened 2026-03-31
Pass --force-refresh to CLI auth token command

mihaimitrea-db:
## 🥞 Stacked PR
Use this [link](https://github.com/databricks/databricks-sdk-java/pull/752/files) to review incremental changes.
- [stack/cli-force-refresh](https://github.com/databricks/databricks-sdk-java/pull/751) [[Files changed](https://github.com/databricks/databricks-sdk-java/pull/751/files)]
  - [**stack/cli-attempt-chain**](https://github.com/databricks/databricks-sdk-java/pull/752) [[Files changed](https://github.com/databricks/databricks-sdk-java/pull/752/files)]

---------
## Summary

Append `--force-refresh` to the `databricks auth token` command when the installed CLI is >= v0.296.0, so the CLI bypasses its internal token cache and hands the SDK a freshly minted token every time.

Mirrors [databricks/databricks-sdk-go#1628](https://github.com/databricks/databricks-sdk-go/pull/1628) and [databricks/databricks-sdk-py#1378](https://github.com/databricks/databricks-sdk-py/pull/1378). Requires the version-detection infrastructure from the parent PR [#751](https://github.com/databricks/databricks-sdk-java/pull/751).

## Why

The SDK already manages its own token caching via `CachedTokenSource`. When the SDK decides it needs a new token and shells out to `databricks auth token`, the CLI may return a token from *its* own cache that is about to expire (or that has already expired from the SDK's perspective). That produces unnecessary refresh failures and retry loops on top of a value that the SDK was confident was fresh.

The CLI added `--force-refresh` in [databricks/cli#4767](https://github.com/databricks/cli/pull/4767) (motivated by [databricks/cli#4564](https://github.com/databricks/cli/issues/4564)) specifically to let callers bypass the CLI's cache. With the version-detection infrastructure from the parent PR already in place, opting in is a one-constant, one-branch change.

### Interface changes

None. `CliTokenSource` is not part of the public API surface.

### Behavioral changes

- `databricks auth token` invocations now end with `--force-refresh` whenever the detected CLI is >= v0.296.0. Callers on older CLIs see no change.
- On older CLIs, a `WARNING` is logged: `"Databricks CLI <ver> does not support --force-refresh (requires >= v0.296.0). The CLI's token cache may provide stale tokens."`

`AzureCliCredentialsProvider` is unaffected — it does not pass through `DatabricksCliCredentialsProvider` and does not opt into version-gated flag selection.

### Internal changes

- New constant `DatabricksCliCredentialsProvider.CLI_VERSION_FOR_FORCE_REFRESH = new DatabricksCliVersion(0, 296, 0)`.
- `buildCliCommand` is split into two helpers, matching the shape the Go and Python SDKs settled on after the same PR there:
  - **`buildCoreCliCommand(cliPath, config, version)`** — holds the existing profile-vs-host decision (moved out of `buildCliCommand`).
  - **`buildCliCommand(cliPath, config, version)`** — now a thin wrapper that calls `buildCoreCliCommand`, appends `--force-refresh` when `version.atLeast(CLI_VERSION_FOR_FORCE_REFRESH)`, and otherwise logs the unsupported-version `WARNING`.

Future version-gated flags slot into the same pattern: add a `CLI_VERSION_FOR_<flag>` constant and an `if version.atLeast(...)` block in `buildCliCommand`.

## How is this tested?

Additional `testBuildCliCommand` parameterized cases in `DatabricksCliCredentialsProviderTest` cover the full matrix:

- `--host` + v0.296.0 → appends `--force-refresh`.
- account `--host` + v0.296.0 → appends `--force-refresh`.
- `--profile` + v0.296.0 → `--profile` + `--force-refresh`.
- `--profile` + v0.207.1 → `--profile` only (too old for `--force-refresh`).
- `--host` + v0.295.0 → `--host` only (too old for `--force-refresh`).
- unknown version (detection failure) → `--host` only, no `--force-refresh`.
- dev build → `--host` only, no `--force-refresh`.

All parent-PR tests continue to pass unchanged.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
@@ -69,6 +69,17 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
     return cmd;
   }
 
+  List<String> buildProfileArgs(String cliPath, DatabricksConfig config) {
+    return new ArrayList<>(
+        Arrays.asList(cliPath, "auth", "token", "--profile", config.getProfile()));
+  }
+
+  private static List<String> withForceRefresh(List<String> cmd) {
+    List<String> forceCmd = new ArrayList<>(cmd);
+    forceCmd.add("--force-refresh");
+    return forceCmd;
+  }
+
   private CliTokenSource getDatabricksCliTokenSource(DatabricksConfig config) {
     String cliPath = config.getDatabricksCliPath();
     if (cliPath == null) {
@@ -81,23 +92,27 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
 
     List<String> cmd;
     List<String> fallbackCmd = null;
+    List<String> secondFallbackCmd = null;
 
     if (config.getProfile() != null) {
-      // When profile is set, use --profile as the primary command.
-      // The profile contains the full config (host, account_id, etc.).
-      cmd =
-          new ArrayList<>(
-              Arrays.asList(cliPath, "auth", "token", "--profile", config.getProfile()));
-      // Build a --host fallback for older CLIs that don't support --profile.
+      List<String> profileArgs = buildProfileArgs(cliPath, config);
+      cmd = withForceRefresh(profileArgs);
+      fallbackCmd = profileArgs;
       if (config.getHost() != null) {
-        fallbackCmd = buildHostArgs(cliPath, config);
+        secondFallbackCmd = buildHostArgs(cliPath, config);
       }
     } else {
       cmd = buildHostArgs(cliPath, config);
     }
 
     return new CliTokenSource(
-        cmd, "token_type", "access_token", "expiry", config.getEnv(), fallbackCmd);
+        cmd,
+        "token_type",
+        "access_token",
+        "expiry",
+        config.getEnv(),
+        fallbackCmd,
+        secondFallbackCmd);
   }
 
   @Override
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
@@ -35,6 +35,10 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
   // --profile support added in CLI v0.207.1: https://github.com/databricks/cli/pull/855
   static final DatabricksCliVersion CLI_VERSION_FOR_PROFILE = new DatabricksCliVersion(0, 207, 1);
 
+  // --force-refresh support added in CLI v0.296.0: https://github.com/databricks/cli/pull/4767
+  static final DatabricksCliVersion CLI_VERSION_FOR_FORCE_REFRESH =
+      new DatabricksCliVersion(0, 296, 0);
+
   // 5-second cap on `databricks version` so a hung CLI (slow first-run scan, antivirus, blocked
   // stdin) does not wedge SDK init indefinitely.
   private static final long VERSION_PROBE_TIMEOUT_SECONDS = 5;
@@ -161,12 +165,40 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
   }
 
   /**
-   * Builds the {@code auth token} command for the given CLI version.
+   * Builds the full {@code auth token} command, including capability-gated flags.
    *
-   * <p>Falls back to {@code --host} when {@code --profile} is either not configured or not
-   * supported by the installed CLI.
+   * <p>Delegates the profile/host decision to {@link #buildCoreCliCommand} and appends {@code
+   * --force-refresh} when the installed CLI supports it.
    */
   List<String> buildCliCommand(
+      String cliPath, DatabricksConfig config, DatabricksCliVersion version) {
+    List<String> cmd = buildCoreCliCommand(cliPath, config, version);
+    if (version.atLeast(CLI_VERSION_FOR_FORCE_REFRESH)) {
+      cmd.add("--force-refresh");
+    } else if (version.isDefaultDevBuild()) {
+      // Dev build — getCliVersion already emitted the dev-build INFO at the probe site.
+    } else if (version.equals(DatabricksCliVersion.UNKNOWN)) {
+      // We didn't actually prove the CLI lacks --force-refresh; we just failed to confirm it.
+      LOG.warn(
+          "Could not confirm --force-refresh support for Databricks CLI {} (requires >= {}). "
+              + "The CLI's token cache may provide stale tokens.",
+          version,
+          CLI_VERSION_FOR_FORCE_REFRESH);
+    } else {
+      LOG.warn(
+          "Databricks CLI {} does not support --force-refresh (requires >= {}). "
+              + "The CLI's token cache may provide stale tokens.",
+          version,
+          CLI_VERSION_FOR_FORCE_REFRESH);
+    }
+    return cmd;
+  }
+
+  /**
+   * Builds the base {@code auth token} command without capability-gated flags. Falls back to {@code
+   * --host} when {@code --profile} is either not configured or not supported by the installed CLI.
+   */
+  List<String> buildCoreCliCommand(
       String cliPath, DatabricksConfig config, DatabricksCliVersion version) {
     if (config.getProfile() == null) {
       return buildHostArgs(cliPath, config);
@@ -176,14 +208,7 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
     // do not support it. Only use --profile in CLI versions known to support it in `auth token`.
     if (!version.atLeast(CLI_VERSION_FOR_PROFILE)) {
       if (version.isDefaultDevBuild()) {
-        // A default-marker dev build has no injected version, so every feature gate fails.
-        // Surface an informational hint so users know why their feature flags aren't taking
-        // effect.
-        LOG.info(
-            "Databricks CLI {} is a development build; feature detection will use conservative "
-                + "fallbacks. Rebuild the CLI with an explicit version to enable capability-based "
-                + "flag selection.",
-            version);
+        // Dev build — getCliVersion already emitted the dev-build INFO at the probe site.
       } else if (version.equals(DatabricksCliVersion.UNKNOWN)) {
         LOG.warn(
             "Could not confirm --profile support for Databricks CLI {} (requires >= {}). "
@@ -254,11 +279,21 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
                 LOG.warn(
                     "Failed to detect Databricks CLI version: {}. "
                         + "Falling back to conservative flag set.",
-                    e.getMessage(),
-                    e);
+                    e.getMessage());
+                LOG.debug("CLI version probe failure stack:", e);
                 return DatabricksCliVersion.UNKNOWN;
               }
             });
+    if (version.isDefaultDevBuild()) {
+      // A default-marker dev build has no injected version, so every feature gate falls back to
+      // the conservative path. Emit once at the probe site so callers see the explanation even
+      // on the host-only code path that never consults a per-flag gate.
+      LOG.info(
+          "Databricks CLI {} is a development build; feature detection will use conservative "
+              + "fallbacks. Rebuild the CLI with an explicit version to enable capability-based "
+              + "flag selection.",
+          version);
+    }
     // Don't cache UNKNOWN: a transient probe failure or a parseable-but-malformed payload
     // would otherwise pin every later token source to the conservative fallback for the rest
     // of the process lifetime. Strip it after computeIfAbsent so the next call re-probes.
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
@@ -80,6 +80,40 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
     return forceCmd;
   }
 
+  List<CliTokenSource.CliCommand> buildCommands(String cliPath, DatabricksConfig config) {
+    List<CliTokenSource.CliCommand> commands = new ArrayList<>();
+
+    boolean hasProfile = config.getProfile() != null;
+    boolean hasHost = config.getHost() != null;
+
+    if (hasProfile) {
+      List<String> profileCmd = buildProfileArgs(cliPath, config);
+
+      commands.add(
+          new CliTokenSource.CliCommand(
+              withForceRefresh(profileCmd),
+              Arrays.asList("--force-refresh", "--profile"),
+              "Databricks CLI does not support --force-refresh flag. "
+                  + "Falling back to regular token fetch. "
+                  + "Please upgrade your CLI to the latest version."));
+
+      commands.add(
+          new CliTokenSource.CliCommand(
+              profileCmd,
+              Collections.singletonList("--profile"),
+              "Databricks CLI does not support --profile flag. Falling back to --host. "
+                  + "Please upgrade your CLI to the latest version."));
+    }
+
+    if (hasHost) {
+      commands.add(
+          new CliTokenSource.CliCommand(
+              buildHostArgs(cliPath, config), Collections.emptyList(), null));
+    }
+
+    return commands;
+  }
+
   private CliTokenSource getDatabricksCliTokenSource(DatabricksConfig config) {
     String cliPath = config.getDatabricksCliPath();
     if (cliPath == null) {
@@ -90,29 +124,8 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
       return null;
     }
 
-    List<String> cmd;
-    List<String> fallbackCmd = null;
-    List<String> secondFallbackCmd = null;
-
-    if (config.getProfile() != null) {
-      List<String> profileArgs = buildProfileArgs(cliPath, config);
-      cmd = withForceRefresh(profileArgs);
-      fallbackCmd = profileArgs;
-      if (config.getHost() != null) {
-        secondFallbackCmd = buildHostArgs(cliPath, config);
-      }
-    } else {
-      cmd = buildHostArgs(cliPath, config);
-    }
-
-    return new CliTokenSource(
-        cmd,
-        "token_type",
-        "access_token",
-        "expiry",
-        config.getEnv(),
-        fallbackCmd,
-        secondFallbackCmd);
+    return CliTokenSource.fromCommands(
+        buildCommands(cliPath, config), "token_type", "access_token", "expiry", config.getEnv());
   }
 
   @Override
--- a/databricks-sdk-java/src/test/java/com/databricks/sdk/core/DatabricksCliCredentialsProviderTest.java
+++ b/databricks-sdk-java/src/test/java/com/databricks/sdk/core/DatabricksCliCredentialsProviderTest.java
@@ -3,6 +3,7 @@ package com.databricks.sdk.core;
 import static org.junit.jupiter.api.Assertions.*;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import org.junit.jupiter.api.Test;
 
@@ -14,6 +15,14 @@ class DatabricksCliCredentialsProviderTest {
   private static final String UNIFIED_HOST = "https://unified.databricks.com";
   private static final String ACCOUNT_ID = "test-account-123";
   private static final String WORKSPACE_ID = "987654321";
+
+  private static final String FORCE_REFRESH_FALLBACK_MSG =
+      "Databricks CLI does not support --force-refresh flag. "
+          + "Falling back to regular token fetch. "
+          + "Please upgrade your CLI to the latest version.";
+  private static final String PROFILE_FALLBACK_MSG =
+      "Databricks CLI does not support --profile flag. Falling back to --host. "
+          + "Please upgrade your CLI to the latest version.";
 
   private final DatabricksCliCredentialsProvider provider = new DatabricksCliCredentialsProvider();
 
@@ -149,5 +158,80 @@ class DatabricksCliCredentialsProviderTest {
     List<String> cmd = provider.buildProfileArgs(CLI_PATH, config);
 
     assertEquals(Arrays.asList(CLI_PATH, "auth", "token", "--profile", "my-profile"), cmd);
+  }
+
+  // ---- Command chain construction tests ----
+
+  @Test
+  void testBuildAttempts_WithProfileAndHost() {
+    DatabricksConfig config = new DatabricksConfig().setHost(HOST).setProfile("my-profile");
+
+    List<CliTokenSource.CliCommand> commands = provider.buildCommands(CLI_PATH, config);
+
+    assertEquals(3, commands.size());
+
+    assertEquals(
+        Arrays.asList(CLI_PATH, "auth", "token", "--profile", "my-profile", "--force-refresh"),
+        commands.get(0).cmd);
+    assertEquals(Arrays.asList("--force-refresh", "--profile"), commands.get(0).usedFlags);
+    assertEquals(FORCE_REFRESH_FALLBACK_MSG, commands.get(0).fallbackMessage);
+
+    assertEquals(
+        Arrays.asList(CLI_PATH, "auth", "token", "--profile", "my-profile"), commands.get(1).cmd);
+    assertEquals(Collections.singletonList("--profile"), commands.get(1).usedFlags);
+    assertEquals(PROFILE_FALLBACK_MSG, commands.get(1).fallbackMessage);
+
+    assertEquals(Arrays.asList(CLI_PATH, "auth", "token", "--host", HOST), commands.get(2).cmd);
+    assertEquals(Collections.emptyList(), commands.get(2).usedFlags);
+    assertNull(commands.get(2).fallbackMessage);
+  }
+
+  @Test
+  void testBuildAttempts_WithProfileOnly() {
+    DatabricksConfig config = new DatabricksConfig().setProfile("my-profile");
+
+    List<CliTokenSource.CliCommand> commands = provider.buildCommands(CLI_PATH, config);
+
+    assertEquals(2, commands.size());
+
+    assertEquals(
+        Arrays.asList(CLI_PATH, "auth", "token", "--profile", "my-profile", "--force-refresh"),
+        commands.get(0).cmd);
+    assertEquals(Arrays.asList("--force-refresh", "--profile"), commands.get(0).usedFlags);
+    assertEquals(FORCE_REFRESH_FALLBACK_MSG, commands.get(0).fallbackMessage);
+
+    assertEquals(
+        Arrays.asList(CLI_PATH, "auth", "token", "--profile", "my-profile"), commands.get(1).cmd);
+    assertEquals(Collections.singletonList("--profile"), commands.get(1).usedFlags);
+    assertEquals(PROFILE_FALLBACK_MSG, commands.get(1).fallbackMessage);
+  }
+
+  @Test
+  void testBuildAttempts_WithHostOnly() {
+    DatabricksConfig config = new DatabricksConfig().setHost(HOST);
+
+    List<CliTokenSource.CliCommand> commands = provider.buildCommands(CLI_PATH, config);
+
+    assertEquals(1, commands.size());
+
+    assertEquals(Arrays.asList(CLI_PATH, "auth", "token", "--host", HOST), commands.get(0).cmd);
+    assertEquals(Collections.emptyList(), commands.get(0).usedFlags);
+    assertNull(commands.get(0).fallbackMessage);
+  }
+
+  @Test
+  void testBuildAttempts_WithAccountHost() {
+    DatabricksConfig config = new DatabricksConfig().setHost(ACCOUNT_HOST).setAccountId(ACCOUNT_ID);
+
+    List<CliTokenSource.CliCommand> commands = provider.buildCommands(CLI_PATH, config);
+
+    assertEquals(1, commands.size());
+
+    assertEquals(
+        Arrays.asList(
+            CLI_PATH, "auth", "token", "--host", ACCOUNT_HOST, "--account-id", ACCOUNT_ID),
+        commands.get(0).cmd);
+    assertEquals(Collections.emptyList(), commands.get(0).usedFlags);
+    assertNull(commands.get(0).fallbackMessage);
   }
 }
--- a/databricks-sdk-java/src/test/java/com/databricks/sdk/core/DatabricksCliCredentialsProviderTest.java
+++ b/databricks-sdk-java/src/test/java/com/databricks/sdk/core/DatabricksCliCredentialsProviderTest.java
@@ -2,8 +2,8 @@ package com.databricks.sdk.core;
 
 import static org.junit.jupiter.api.Assertions.*;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.mockConstruction;
@@ -16,7 +16,6 @@ import com.databricks.sdk.core.utils.Environment;
 import com.databricks.sdk.core.utils.OSUtilities;
 import com.databricks.sdk.core.utils.OSUtils;
 import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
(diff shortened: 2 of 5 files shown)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+   */
+  DatabricksCliVersion getCliVersion(String cliPath, Environment env) {
+    DatabricksCliVersion cached = VERSION_CACHE.get(cliPath);
+    if (cached != null) {
+      return cached;
+    }
+
+    try {
+      DatabricksCliVersion version = probeCliVersion(cliPath, env);
+      VERSION_CACHE.put(cliPath, version);
```
simonfaltum: **[Major] Parse failures cached as successful `UNKNOWN` - violates the "don't cache failures" contract.**

`probeCliVersion` runs the subprocess successfully (exit 0), then returns `parseCliVersion(stdout)`. But `parseCliVersion` returns `DatabricksCliVersion.UNKNOWN` on malformed JSON or missing `Major`/`Minor`/`Patch`. Because the try/catch at line 265 only catches exceptions, the `UNKNOWN` value gets stored in `VERSION_CACHE` here. That pins every future `getCliVersion(cliPath)` call to the conservative fallback for the rest of the JVM lifetime, exactly the regression the "failures are not cached" invariant (stated in the cache comment and the PR description) was designed to prevent. A transiently corrupt CLI response permanently disables `--force-refresh` (and any future gated flag).

Suggestion: treat parse failure as a probe failure. Either have `parseCliVersion` throw on malformed input (letting the outer catch at line 265 return UNKNOWN without caching), or only cache when `!version.equals(UNKNOWN)`.

Found by: Cursor (cross-review).

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+      JsonNode node = MAPPER.readTree(output);
+      JsonNode major = node.get("Major");
+      JsonNode minor = node.get("Minor");
+      JsonNode patch = node.get("Patch");
+      if (major == null || minor == null || patch == null) {
+        LOG.debug(
+            "Failed to parse Databricks CLI version: missing Major/Minor/Patch in {}", output);
+        return DatabricksCliVersion.UNKNOWN;
+      }
+      return new DatabricksCliVersion(major.asInt(), minor.asInt(), patch.asInt());
```
simonfaltum, line 397: **[Major] `asInt()` silently coerces non-integer fields to 0, can enable flags that should be gated off.**

`JsonNode.asInt()` returns 0 for any non-numeric node (null, string, array). Input like `{"Major":"v0","Minor":296,"Patch":0}` parses here as `v0.296.0` and enables `--force-refresh`. Input like `{"Major":0,"Minor":"bad","Patch":0}` parses as `(0,0,0)` = dev build, not UNKNOWN, which also matters because dev build vs UNKNOWN log at different levels. This violates the "bad output -> conservative command" invariant stated in the javadoc, and paired with the cached-UNKNOWN finding above undermines fail-closed behavior in two directions at once.

Suggestion: check `major.isIntegralNumber()` / `isInt()` on each field and return UNKNOWN if any is not a numeric node. Consider rejecting negative values too (they'd parse cleanly today but aren't meaningful).

Found by: Cursor (raised to Major in cross-review) + Isaac (Nit in R1).

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+  DatabricksCliVersion probeCliVersion(String cliPath, Environment env) throws IOException {
+    List<String> versionArgs = Arrays.asList(cliPath, "version", "--output", "json");
+    List<String> cmd = OSUtils.get(env).getCliExecutableCommand(versionArgs);
+
+    ProcessBuilder pb = new ProcessBuilder(cmd);
+    pb.environment().putAll(env.getEnv());
+    Process process = pb.start();
+
+    try {
+      if (!process.waitFor(VERSION_PROBE_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
```
simonfaltum: **[Major] Subprocess stream draining: latent pipe-buffer deadlock.**

`process.waitFor(5s)` is called here before either stream is read. If the child ever writes more than the OS pipe buffer (~4-64 KB depending on platform) to stdout or stderr, the child blocks on `write()`, the parent blocks in `waitFor`, and the probe spuriously times out at 5s. This is a textbook Java `ProcessBuilder` gotcha. Today `databricks version --output json` emits ~100 bytes so it never fires, but the moment a future CLI release adds a deprecation banner, proxy diagnostic, or telemetry line to stderr, version detection silently breaks and every caller eats a 5-second timeout per `configure()` with no useful diagnostic.

Suggestion: either (a) `pb.redirectErrorStream(true)` and drain the combined stream on a background thread while `waitFor` runs, or (b) drain stdout and stderr on two background threads before/during `waitFor`. The idiomatic pattern:

```java
pb.redirectErrorStream(true);
Process p = pb.start();
p.getOutputStream().close();  // see separate stdin finding
CompletableFuture<String> out = CompletableFuture.supplyAsync(() -> readAll(p.getInputStream()));
if (!p.waitFor(VERSION_PROBE_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
  p.destroyForcibly();
  throw new IOException("timed out ...");
}
return parseCliVersion(out.get(1, TimeUnit.SECONDS));
```

Found by: Cursor (Major) + Isaac (Suggestion, leaned toward Major in cross-review).

databricks-sdk-java/src/test/java/com/databricks/sdk/core/DatabricksCliCredentialsProviderTest.java
```diff
@@ -37,12 +44,125 @@ void testBuildHostArgs_AccountHost() {
...
+        Arguments.of(
+            "dev build — falls back to --host, no force-refresh",
+            new DatabricksConfig().setProfile(PROFILE).setHost(HOST),
+            new DatabricksCliVersion(0, 0, 0),
+            Arrays.asList(CLI_PATH, "auth", "token", "--host", HOST)));
+  }
+
+  @ParameterizedTest(name = "{0}")
+  @MethodSource("buildCliCommandCases")
+  void testBuildCliCommand(
```
simonfaltum, line 171: **[Gap (Major)] No tests for `getCliVersion` caching or `probeCliVersion` subprocess.**

The parameterized matrix below is well-structured for `buildCliCommand` (the pure formatter), but `getCliVersion` and `probeCliVersion` have zero direct coverage. The PR description and the `getCliVersion` javadoc both promise: *"Successful results are cached per `cliPath` for the process lifetime; failures are not cached and will be retried on the next call."* That is the central correctness claim of the version-detection infrastructure, caching a transient failure would pin the SDK to the conservative fallback for the JVM lifetime, which is the exact regression this design was meant to avoid. It is currently untested.

Subprocess behavior, argument construction (including the OS wrapping via `OSUtils.get(env).getCliExecutableCommand(...)`), env propagation, non-zero exit + stderr capture, timeout with `destroyForcibly`, and parse-failure handling, is also untested. These are the parts most likely to break across OSes.

Suggestion: using `MockedConstruction<ProcessBuilder>` (the pattern already established in `CliTokenSourceTest`), add:

1. Successful probe + second call -> only 1 ProcessBuilder constructed (cache hit).
2. Failed probe (non-zero exit / timeout / malformed JSON output) + successful probe -> 2 ProcessBuilders constructed (failure not cached, second result wins).
3. Two distinct `cliPath` values maintain separate cache entries.
4. Timeout path calls `destroyForcibly` and throws `IOException` with the timeout message.
5. Non-zero exit surfaces stderr in the `IOException` message.
6. `getCliExecutableCommand` is invoked with `[cliPath, "version", "--output", "json"]` (so the OS wrapper is actually applied).

The static `VERSION_CACHE` will need a reset hook (`@VisibleForTesting static void clearVersionCache()`) called from `@BeforeEach` to keep tests order-independent.

Found by: Both.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+   * Runs {@code databricks version --output json} and returns the parsed {@link
+   * DatabricksCliVersion}.
+   */
+  DatabricksCliVersion probeCliVersion(String cliPath, Environment env) throws IOException {
+    List<String> versionArgs = Arrays.asList(cliPath, "version", "--output", "json");
+    List<String> cmd = OSUtils.get(env).getCliExecutableCommand(versionArgs);
+
+    ProcessBuilder pb = new ProcessBuilder(cmd);
+    pb.environment().putAll(env.getEnv());
+    Process process = pb.start();
```
simonfaltum, line 324: **[Nit] Child stdin is never closed, interactive prompts would hang until timeout.**

`ProcessBuilder` leaves the child's stdin as an open pipe to the parent. If the CLI ever reads from stdin (future interactive prompt, TTY-triggered behavior, some flag we didn't pass), the child blocks on `read()` and we eat the full 5-second timeout for every `configure()` call. Today `databricks version` doesn't read stdin, so this doesn't fire in practice, but it's a one-line fix to harden against it.

Suggestion:
```java
Process process = pb.start();
process.getOutputStream().close();
```
Or, before `start()`: `pb.redirectInput(ProcessBuilder.Redirect.from(new File(isWindows ? "NUL" : "/dev/null")))`.

Found by: Cursor.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+    } catch (JsonProcessingException e) {
+      LOG.debug(
+          "Failed to parse Databricks CLI version from output: {} ({})", output, e.getMessage());
+      return DatabricksCliVersion.UNKNOWN;
+    }
+  }
+
+  private static String readStream(InputStream stream) throws IOException {
+    return new String(IOUtils.toByteArray(stream), StandardCharsets.UTF_8);
+  }
```
simonfaltum, line 411: **[Nit] `readStream` leaks the `InputStream` wrapper.**

`IOUtils.toByteArray(stream)` does not close the stream, and the two callers at lines 299 and 304 don't close either. Each probe leaks 1-2 stream wrappers. The underlying OS-level FDs are released when the `Process` exits, but the Java wrapper hangs around until GC. The successful-probe cache keeps this to one leak per `cliPath` in practice, so this is low impact, but it's a one-line fix.

Suggestion:
```java
private static String readStream(InputStream stream) throws IOException {
  try (InputStream in = stream) {
    return new String(IOUtils.toByteArray(in), StandardCharsets.UTF_8);
  }
}
```

Found by: Cursor (R1) + Isaac (cross-review).

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+                + "Falling back to --host.",
+            version,
+            CLI_VERSION_FOR_PROFILE);
+      } else {
+        LOG.warn(
+            "Databricks CLI {} does not support --profile (requires >= {}). Falling back to --host.",
+            version,
+            CLI_VERSION_FOR_PROFILE);
+      }
+      return buildHostArgs(cliPath, config);
```
simonfaltum, line 225: **[Nit] `buildCoreCliCommand` can build `["--host", null]` via this package-private helper.**

When `config.getProfile() != null`, `config.getHost() == null`, and the CLI doesn't support `--profile` (old / UNKNOWN / dev-build), this line unconditionally falls through to `buildHostArgs`, which produces `["databricks", "auth", "token", "--host", null]`. `ProcessBuilder.start()` then throws NPE deep in `ProcessImpl` with no useful diagnostic.

**In the production flow this is not reachable today**: `configure()` returns `null` early when `config.getHost() == null` (see the top of `configure()`), so `getDatabricksCliTokenSource` never runs with a null host. But the old code path guarded this same condition *inside* `getDatabricksCliTokenSource`; removing that guard while relying on the outer `configure()` guard weakens the local invariant. If a future refactor loosens the `configure()` early return, if `buildCoreCliCommand` is called from a new path, or if a test calls this package-private helper directly with a profile-only config, you get a cryptic NPE instead of a clear error.

Suggestion: either add `if (config.getHost() == null) throw new DatabricksException("...");` inside `buildCoreCliCommand` before the `buildHostArgs` fallback, or add a javadoc note pinning the precondition. Add a test for `profile + null host + old CLI` either way.

Found by: Both (Isaac Major, Cursor Gap Minor; downgraded to Nit after cross-review confirmed the production path is gated by `configure()`'s early return).

simonfaltum (commented):
## Swarm Review Summary (Isaac + Cursor, 2 rounds)

**Verdict: Not ready yet.** 4 Major, 11 Nit, 4 Suggestion. No Critical.

The design mirrors the Go/Python SDK implementations well, `DatabricksCliVersion` is clean, and the `buildCliCommand` test matrix is solid. The concerns are concentrated in `probeCliVersion` + `parseCliVersion` + `getCliVersion`:

### The four Majors
1. **Parse failures are cached as successful `UNKNOWN`** (line 263). Violates the "don't cache failures" contract, a transiently corrupt CLI response pins the SDK to the conservative fallback for the rest of the JVM lifetime.
2. **`asInt()` silently coerces non-integer fields to 0** (line 329). `{"Major":"v0","Minor":296,"Patch":0}` parses as `v0.296.0` and *enables* `--force-refresh`. Strictly worse failure mode than UNKNOWN.
3. **Subprocess stream draining: latent pipe-buffer deadlock** (line 286). `waitFor` before draining stdout/stderr is the textbook Java `ProcessBuilder` gotcha. Works today because output is ~100 bytes; breaks the first time a CLI release adds a stderr banner.
4. **No tests for `getCliVersion` caching or `probeCliVersion` subprocess paths.** The central correctness claim of this infrastructure, "cache successes, retry failures", is completely untested. Subprocess argument construction, OS wrapping, exit handling, timeout, and parse-failure handling also have zero coverage.

### What went well
- `DatabricksCliVersion` is clean, well-commented, with correct `compareTo`/`equals`/`hashCode`, and handles sentinels in `toString`.
- Cache-by-cliPath (not by `DatabricksConfig`) is the right key.
- "Don't cache failures" is the right policy and is explicitly documented, the bug is that the implementation doesn't match the policy for the parse-failure path.
- Explicit dev-build `(0, 0, 0)` sentinel with `isDefaultDevBuild()` is thoughtful.
- Removing `fallbackCmd` from `CliTokenSource` is a real simplification.

### Reviewer alignment
Isaac and Cursor independently converged on: the subprocess deadlock, test-coverage gap, log noise (3 WARNs per failed probe), package-private constants, `VERSION_CACHE` reset hook, and concurrent-probe dedup. That's strong signal on the areas above.

Isaac found the `--host null` regression in `buildCoreCliCommand`; Cursor verified it is gated by `configure()`'s early-return in the production path, so it's downgraded to Nit as a local-invariant concern (still worth a guard or a javadoc note).

Cursor found the cached-UNKNOWN bug and the `asInt()` coercion issue, the two most consequential correctness findings, during cross-review.

Full synthesis: inline comments below.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+    if (cached != null) {
+      return cached;
+    }
+
+    try {
+      DatabricksCliVersion version = probeCliVersion(cliPath, env);
+      VERSION_CACHE.put(cliPath, version);
+      return version;
+    } catch (Exception e) {
+      LOG.warn(
```
simonfaltum: **[Nit] Single failed probe produces 3 WARN logs per `configure()`.**

A single failed probe with a profile-set config fires:
1. This WARN: "Failed to detect Databricks CLI version..."
2. WARN at line 213: "Could not confirm --profile support..."
3. WARN at line 183: "Could not confirm --force-refresh support..."

All three point at the same root cause (UNKNOWN version). For customers instantiating many `DatabricksConfig`s (Spark-job-per-session pattern, for example), this is a lot of noise that buries the actual probe-failure diagnostic.

Suggestion: log the probe failure once at WARN here with the full underlying message. Demote the two per-flag "could not confirm" paths to DEBUG when `version.equals(UNKNOWN)` (the single WARN at the probe site already explains why). Keep WARN for the "confirmed old CLI" branches, those are actionable (upgrade).

Found by: Cursor (R1) + Isaac (cross-review).

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+    if (version.atLeast(CLI_VERSION_FOR_FORCE_REFRESH)) {
+      cmd.add("--force-refresh");
+    } else if (version.equals(DatabricksCliVersion.UNKNOWN) || version.isDefaultDevBuild()) {
+      // Detection failed or no version metadata — we can't prove the CLI lacks --force-refresh,
+      // just failed to confirm it. The version probe already logged the underlying cause.
+      LOG.warn(
+          "Could not confirm --force-refresh support for Databricks CLI {} (requires >= {}). "
+              + "The CLI's token cache may provide stale tokens.",
+          version,
+          CLI_VERSION_FOR_FORCE_REFRESH);
```
simonfaltum: **[Nit] Dev build produces 3 logs per `configure()`.**

Dev build `(0, 0, 0)` always compares less than `CLI_VERSION_FOR_FORCE_REFRESH = (0, 296, 0)` and hits this else branch. Today that means: INFO in `resolveCliCommand` ("dev build, conservative fallbacks"), then this WARN, then the WARN in `buildCoreCliCommand` if a profile is set. Three logs for one condition.

Suggestion: keep the single `resolveCliCommand` INFO and demote the per-flag branches to DEBUG when `version.isDefaultDevBuild()`. Also worth splitting the `UNKNOWN || isDefaultDevBuild()` branch, they're conceptually different (one is "probe failed", the other is "probe succeeded but CLI was built without version metadata") and share a log message that fits neither perfectly.

Found by: Cursor.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -22,6 +29,26 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
...
+      new DatabricksCliVersion(0, 296, 0);
+
+  // 5-second cap on `databricks version` so a hung CLI (slow first-run scan, antivirus, blocked
+  // stdin) does not wedge SDK init indefinitely.
+  private static final long VERSION_PROBE_TIMEOUT_SECONDS = 5;
+
+  // Successful version probes keyed by cliPath. Failures are deliberately not cached, so a
+  // transient error (timeout, AV scan) does not pin every later token source to the conservative
+  // fallback for the rest of the process lifetime.
+  private static final Map<String, DatabricksCliVersion> VERSION_CACHE = new ConcurrentHashMap<>();
```
simonfaltum, line 50: **[Nit] Static `VERSION_CACHE` has no reset hook.**

Two consequences:
1. Once the missing cache tests (see R2 Major finding) are added, results leak across test methods -> order-dependent flakes.
2. If anything swaps the CLI binary at the same path mid-process (Homebrew upgrade, CI runner infra, `databricks --upgrade`), the cached version stays stale for the JVM lifetime.

Suggestion: add a `@VisibleForTesting static void clearVersionCache()` and call it from `@BeforeEach` in the new cache tests. Consider whether an instance-scoped cache would be cleaner here, static mirrors the Go/Python SDKs so that's a judgment call.

Found by: Both.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+        process.destroyForcibly();
+        throw new IOException(
+            "timed out after "
+                + VERSION_PROBE_TIMEOUT_SECONDS
+                + "s waiting for `databricks version`");
+      }
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new IOException("interrupted waiting for `databricks version`", e);
+    }
```
simonfaltum, line 333: **[Nit] Process leaks on the `InterruptedException` path.**

When `waitFor` is interrupted, the catch block re-sets the interrupt flag and throws `IOException`, but does not call `process.destroyForcibly()`. The child subprocess keeps running in the background until JVM exit. This is asymmetric with the timeout branch (line 287) which does destroy. Low frequency since interrupts during version probing are rare, but trivially fixable.

Suggestion: `process.destroyForcibly();` (best-effort) before re-throwing in the `InterruptedException` handler.

Found by: Isaac (cross-review).

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+
+    if (process.exitValue() != 0) {
+      String stderr = readStream(process.getErrorStream());
+      throw new IOException(
+          "`databricks version` exited with code " + process.exitValue() + ": " + stderr);
+    }
```
simonfaltum, line 350: **[Nit] On non-zero exit, stdout is never drained, second facet of the deadlock concern.**

When `exitValue() != 0`, only stderr is read; stdout is left undrained. This is a separate facet of the pipe-buffer deadlock concern: even after fixing the success-path deadlock, the failure path still skips stdout entirely. A future CLI that sends help text to stdout on a bad argument (common with Cobra-based CLIs) would deadlock here for the same pipe-buffer reason.

Suggestion: drain both streams unconditionally before checking exit code. This folds naturally into the `redirectErrorStream(true)` or two-thread-drain fix for the R2 Major finding on line 286.

Found by: Isaac (cross-review).

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+    Process process = pb.start();
+
+    try {
+      if (!process.waitFor(VERSION_PROBE_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
+        process.destroyForcibly();
+        throw new IOException(
+            "timed out after "
+                + VERSION_PROBE_TIMEOUT_SECONDS
+                + "s waiting for `databricks version`");
+      }
```
simonfaltum: **[Nit] `destroyForcibly` not awaited; streams never closed on exit paths.**

`destroyForcibly()` is asynchronous on some platforms; without a follow-up `waitFor`, the process handle and its FDs may linger until GC. Streams are never explicitly closed on either the timeout or non-zero-exit path.

Suggestion: `process.destroyForcibly().waitFor(<short>, TimeUnit.SECONDS)` on the timeout path. Wrap the entire probe body in `try { ... } finally { closeQuietly(stdout); closeQuietly(stderr); }` (Apache Commons `IOUtils.closeQuietly` keeps it short).

Found by: Isaac.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -22,6 +29,26 @@ public class DatabricksCliCredentialsProvider implements CredentialsProvider {
...
   private static final ObjectMapper MAPPER = new ObjectMapper();
 
+  // ---- Version detection ----
+
+  // --profile support added in CLI v0.207.1: https://github.com/databricks/cli/pull/855
+  static final DatabricksCliVersion CLI_VERSION_FOR_PROFILE = new DatabricksCliVersion(0, 207, 1);
+
+  // --force-refresh support added in CLI v0.296.0: https://github.com/databricks/cli/pull/4767
+  static final DatabricksCliVersion CLI_VERSION_FOR_FORCE_REFRESH =
+      new DatabricksCliVersion(0, 296, 0);
```
simonfaltum, line 40: **[Nit] `CLI_VERSION_FOR_PROFILE` / `CLI_VERSION_FOR_FORCE_REFRESH` are package-private but unused outside this class.**

The constants are `static final` (package-private) but nothing references them from outside, even the tests hardcode `(0, 207, 1)` / `(0, 296, 0)` rather than importing the constant. That means a typo in either constant won't be caught by the tests. Two options:

1. Make them `private`, simpler, matches current usage.
2. Keep package-private and have the tests reference them (e.g. `DatabricksCliCredentialsProvider.CLI_VERSION_FOR_FORCE_REFRESH`), preferred, because it keeps source constant and test expectations in lockstep.

Found by: Both.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+   *
+   * <p>Takes Major/Minor/Patch from the JSON's pre-parsed numeric fields. The Prerelease field and
+   * the Version string are intentionally ignored: for our feature-gate purposes the base triple is
+   * sufficient, and the (0, 0, 0) case already identifies the default dev build (a CLI built
+   * without version metadata leaves these fields at their zero defaults).
+   *
+   * <p>Returns {@link DatabricksCliVersion#UNKNOWN} on failure so that an unparseable version
+   * disables every feature gate.
+   */
+  static DatabricksCliVersion parseCliVersion(String output) {
```
simonfaltum: **[Nit] Mixed static and instance methods with no clear rule.**

`parseCliVersion` is `static`, while `getCliVersion`, `probeCliVersion`, `buildCliCommand`, `buildCoreCliCommand`, and `buildHostArgs` are instance methods, yet none of them touch instance state (`MAPPER` and `VERSION_CACHE` are both static). The tests reflect the arbitrary split: `provider.buildCliCommand(...)` as instance vs `DatabricksCliCredentialsProvider.parseCliVersion(...)` as static.

Suggestion: make them all `static` (matches their behavior, none read `this`) or all instance. Lean static.

Found by: Isaac.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+          "Could not confirm --force-refresh support for Databricks CLI {} (requires >= {}). "
+              + "The CLI's token cache may provide stale tokens.",
+          version,
+          CLI_VERSION_FOR_FORCE_REFRESH);
+    } else {
+      LOG.warn(
+          "Databricks CLI {} does not support --force-refresh (requires >= {}). "
+              + "The CLI's token cache may provide stale tokens.",
+          version,
+          CLI_VERSION_FOR_FORCE_REFRESH);
```
simonfaltum, line 192: **[Nit] Warning message has no actionable fix for the user.**

"Databricks CLI v0.xxx does not support --force-refresh (requires >= v0.296.0). The CLI's token cache may provide stale tokens." - a user who sees this has no idea that the fix is to upgrade the CLI.

Suggestion: append an upgrade hint, e.g. "Upgrade via `brew upgrade databricks` or https://docs.databricks.com/dev-tools/cli/install.html to silence this warning." Match the phrasing used in the Go SDK's equivalent message for cross-SDK consistency.

Found by: Cursor.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+  DatabricksCliVersion getCliVersion(String cliPath, Environment env) {
+    DatabricksCliVersion cached = VERSION_CACHE.get(cliPath);
+    if (cached != null) {
+      return cached;
+    }
+
+    try {
+      DatabricksCliVersion version = probeCliVersion(cliPath, env);
+      VERSION_CACHE.put(cliPath, version);
+      return version;
```
simonfaltum: **[Suggestion] `getCliVersion` get-then-put is non-atomic; concurrent cold calls double-probe.**

Two threads calling `getCliVersion` for the same `cliPath` on a cold cache will both spawn `databricks version` subprocesses. Not a correctness issue (both threads write the same value), but it wastes a subprocess and a 5s-bounded blocking call. Matters when many `configure()` calls fire simultaneously at startup.

Suggestion: `computeIfAbsent` with care, the failure-not-cached invariant requires not inserting on exception. Cleanest: probe inside the lambda, throw an unchecked sentinel on failure, catch outside and return UNKNOWN without caching. Or use a per-path `synchronized` block around a double-checked lookup.

Low priority, flagging because the cache comment already advertises determinism per path and the behavior today doesn't match that advertisement under concurrency.

Found by: Both.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+    pb.environment().putAll(env.getEnv());
+    Process process = pb.start();
+
+    try {
+      if (!process.waitFor(VERSION_PROBE_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
+        process.destroyForcibly();
+        throw new IOException(
+            "timed out after "
+                + VERSION_PROBE_TIMEOUT_SECONDS
+                + "s waiting for `databricks version`");
```
simonfaltum: **[Suggestion] Capture and surface stderr on the timeout path for diagnostics.**

On timeout, the `IOException` message is "timed out after 5s waiting for `databricks version`" with no clue *why* the CLI hung. Even partial stderr ("Initializing telemetry...", "Locking config dir...", "Checking for updates...") would help users diagnose AV scanners, lock contention, proxy issues, or first-run setup.

Suggestion: best-effort `readStream(process.getErrorStream())` after `destroyForcibly().waitFor(<short>)` and append to the exception message. Wrap in its own try/catch, the stream may already be closed. This folds naturally into the concurrent-drain fix for the Major finding above.

Found by: Isaac.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
simonfaltum: **[Suggestion] Log a DEBUG line on successful version detection.**

When a user files a bug saying "force-refresh isn't being passed", the first diagnostic question is "what version did the probe detect?" Currently there's no way to answer without running under a debugger. One DEBUG line gives us free diagnosability.

Suggestion: `LOG.debug("Detected Databricks CLI {} at {}", version, cliPath);` after the successful `probeCliVersion` call.

Found by: Cursor.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +132,214 @@ public Token getToken() {
...
+          CLI_VERSION_FOR_FORCE_REFRESH);
+    } else {
+      LOG.warn(
+          "Databricks CLI {} does not support --force-refresh (requires >= {}). "
+              + "The CLI's token cache may provide stale tokens.",
+          version,
+          CLI_VERSION_FOR_FORCE_REFRESH);
+    }
+    return cmd;
+  }
```
simonfaltum, line 195: **[Suggestion] Defensive copy before mutating the list from `buildCoreCliCommand`.**

`buildCliCommand` mutates the list returned by `buildCoreCliCommand` via `cmd.add("--force-refresh")`. This works today only because `buildCoreCliCommand` and `buildHostArgs` happen to return `ArrayList`. A future refactor returning `List.of(...)` or `Collections.unmodifiableList(...)` will silently break this with `UnsupportedOperationException` at a call site that looks innocent.

Suggestion: either `List<String> cmd = new ArrayList<>(buildCoreCliCommand(...));` at the top of this method, or document the mutability contract in javadoc on `buildCoreCliCommand`. The copy is more robust for a trivial cost.

Found by: Isaac.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
mihaimitrea-db: Duplicate of comment in previous PR.

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +139,277 @@ public Token getToken() {
...
+            key,
+            k -> {
+              try {
+                return probeCliVersion(cliPath, env);
+              } catch (IOException e) {
+                LOG.warn(
+                    "Failed to detect Databricks CLI version: {}. "
+                        + "Falling back to conservative flag set.",
+                    e.getMessage(),
+                    e);
```
simonfaltum: Nit: passing `e` as a trailing argument to `LOG.warn(...)` makes SLF4J emit a full stack trace at WARN. Combined with the deliberate don't-cache-UNKNOWN policy a few lines down, a host with a persistently broken CLI will re-probe on every `DatabricksConfig` init and re-log the stack each time. Not a blocker, but worth demoting:

```java
LOG.warn(
    "Failed to detect Databricks CLI version: {}. Falling back to conservative flag set.",
    e.getMessage());
LOG.debug("CLI version probe failure stack:", e);
```

databricks-sdk-java/src/main/java/com/databricks/sdk/core/DatabricksCliCredentialsProvider.java
```diff
@@ -151,6 +139,277 @@ public Token getToken() {
...
+   * --force-refresh} when the installed CLI supports it.
+   */
+  List<String> buildCliCommand(
+      String cliPath, DatabricksConfig config, DatabricksCliVersion version) {
+    List<String> cmd = buildCoreCliCommand(cliPath, config, version);
+    if (version.atLeast(CLI_VERSION_FOR_FORCE_REFRESH)) {
+      cmd.add("--force-refresh");
+    } else if (version.isDefaultDevBuild()) {
+      // Dev build — the dev-build INFO emitted by buildCoreCliCommand already explained why
+      // feature detection is disabled; no per-flag noise needed here.
```
simonfaltum: Nit / minor precondition gap: this branch's comment relies on `buildCoreCliCommand` having already emitted the dev-build INFO, but `buildCoreCliCommand` early-returns at `config.getProfile() == null` *before* reaching its dev-build log. So a dev-build CLI + host-only config silently runs the conservative command with no log explaining why `--force-refresh` is absent. Narrow slice of users (CLIs built without version metadata), but a small usability hole and the pattern won't compose cleanly when future capability gates are added.

Consider emitting the dev-build INFO once from `resolveCliCommand` or `getCliVersion` right after the probe when `version.isDefaultDevBuild()`. Each capability gate then becomes a simple `if at-least, append; else if dev/UNKNOWN, swallow; else WARN` with no cross-call assumption.

databricks-sdk-java/src/test/java/com/databricks/sdk/core/DatabricksCliCredentialsProviderTest.java
```diff
@@ -37,12 +78,414 @@ void testBuildHostArgs_AccountHost() {
...
+  }
+
+  // ---- probeCliVersion subprocess tests ----
+
+  /** Builds a mock Process whose merged-output stream returns {@code stdout}. */
+  private static Process mockProcess(String stdout, int exitCode, boolean exited) throws Exception {
+    Process process = mock(Process.class);
+    when(process.getInputStream())
+        .thenReturn(new ByteArrayInputStream(stdout.getBytes(StandardCharsets.UTF_8)));
+    when(process.getOutputStream()).thenReturn(new ByteArrayOutputStream());
```
simonfaltum: Nit: `probeCliVersion` never calls `getOutputStream()` (it reads merged stdout via `getInputStream()` after `redirectErrorStream(true)`), so this stub is dead. Safe to drop.

databricks-sdk-java/src/test/java/com/databricks/sdk/core/DatabricksCliCredentialsProviderTest.java
```diff
@@ -37,12 +78,414 @@ void testBuildHostArgs_AccountHost() {
...
+    OSUtilities osUtils = passthroughOsUtils();
+    Process process =
+        mockProcess("{\"Version\":\"v0.295.0\",\"Major\":0,\"Minor\":295,\"Patch\":0}", 0, true);
+
+    try (MockedStatic<OSUtils> mockedOSUtils = mockStatic(OSUtils.class);
+        MockedConstruction<ProcessBuilder> mocked =
+            mockConstruction(
+                ProcessBuilder.class,
+                (pb, ctx) -> {
+                  when(pb.redirectErrorStream(anyBoolean())).thenReturn(pb);
```
simonfaltum: Nit: production calls `pb.redirectErrorStream(true)` with a literal — the test could use `eq(true)` to lock the merged-stream invariant. `anyBoolean()` would silently pass a future regression that flips it to `false` and re-introduces deadlock risk on large stderr output. Same suggestion applies to lines 453 and 477.

simonfaltum (approved):
Approved. Deep-reviewed end-to-end with two independent reviewers (Isaac + Cursor, two rounds) against the full stack (#751 + #752). Core logic is correct: the version gate works, `--force-refresh` is appended correctly when the CLI supports it, the profile-vs-host fallback is sound, and the test matrix covers the interesting combinations. Sibling Go and Python SDKs converged on the same shape, which is reassuring.

No correctness, security, data-loss, or behavioral-regression concerns. The version-to-PR mappings were verified against the CLI repo: `databricks/cli#855` first lands in v0.207.1 and `databricks/cli#4767` first lands in v0.296.0. Both constants are correct.

A few nits inline, none blocking:
- WARN log on probe failure passes the throwable as a trailing arg, which triggers a full stack trace at WARN. Combined with don't-cache-UNKNOWN, this can be noisy for hosts that construct many configs against a broken CLI. Demote the stack to DEBUG.
- `buildCliCommand`'s dev-build silent branch has a precondition gap (relies on `buildCoreCliCommand` having logged, but the profile=null early-return skips that log). Narrow but worth cleaning up by centralizing the dev-build INFO near the probe.
- Two test-side nits: an unused `getOutputStream()` stub, and `anyBoolean()` matchers that should be `eq(true)` to lock the `redirectErrorStream` invariant.

A few things worth considering in a follow-up but not for this PR:
- Short TTL on cached UNKNOWN so persistent probe failures don't pay the 5s timeout on every init.
- `toRealPath()` for the cache key to share entries across symlinked installs.
- Static `VERSION_CACHE` plus `clearVersionCache()` is a parallel-test hazard if the project ever enables parallel execution; instance field would be safer.
- No tests for the `InterruptedException` path in `probeCliVersion` or for `parseCliVersion(null)` — neither is a runtime bug in the current call paths, but both would catch future refactors.

Full synthesis available on request.

PR #758, opened 2026-04-13
Fix error type for non-JSON error responses

renaudhartert-db:
## Summary

When the server returns a plain-text error response (e.g. `"Invalid Token"` with HTTP 403), the SDK throws `Unknown` instead of `PermissionDenied`.

The root cause: `parseUnknownError` derives an error code by splitting `response.getStatus()` on space, expecting `"403 Forbidden"`. But `CommonsHttpClient` passes `statusLine.getReasonPhrase()`, which is just `"Forbidden"`. The split produces one element, so error code defaults to `"UNKNOWN"`, which matches `ErrorMapper`'s error code mapping and short-circuits the status code mapping that would have correctly produced `PermissionDenied`.

The fix removes the error code derivation entirely. Leaving `errorCode` null lets `AbstractErrorMapper.apply` skip the error code mapping and fall through to the status code mapping (403 -> `PermissionDenied`, 401 -> `Unauthenticated`, etc.). The error message now uses the raw response body instead of appending Jackson parse exception details.

**Behavioral change**: non-JSON error responses now produce typed exceptions based on HTTP status code instead of always producing `Unknown`. The error message no longer contains Jackson deserialization internals.

## Test plan

- [x] New `PlainTextErrorTest` covers: plain-text 403/401/404, HTML `<pre>` extraction, empty body, null body
- [x] Full test suite passes (`mvn test -pl databricks-sdk-java`)


databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/ApiErrors.java
```diff
@@ -116,27 +116,17 @@ private static Optional<ApiErrorBody> parseApiError(Response response) {
     try {
       return Optional.of(MAPPER.readValue(body, ApiErrorBody.class));
     } catch (IOException e) {
-      return Optional.of(parseUnknownError(response, body, e));
+      return Optional.of(parseUnknownError(body));
```
mihaimitrea-db, line 119: Would leaving `errorCode` null cause NPE if someone downstream calls `e.getErrorCode().equals()`:

ClustersExt.java: L221-222:
```java
} catch (DatabricksError e) {
  if (e.getErrorCode().equals(INVALID_STATE)) {
````

Could we add a dummy value that would always make `AbstractErrorMapper.apply` skip the error code mapping while avoiding NPE?

Comment (renaudhartert-db) [reply databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/ApiErrors.java:119]: Good catch, done.

File: databricks-sdk-java/src/main/java/com/databricks/sdk/core/error/ApiErrors.java [hunk 2]
```diff
@@ -116,27 +116,17 @@ private static Optional<ApiErrorBody> parseApiError(Response response) {
...
-      errorBody.setErrorCode("UNKNOWN");
-    } else {
-      String errorCode = statusParts[1].replaceAll("^[ .]+|[ .]+$", "");
-      errorBody.setErrorCode(errorCode.replaceAll(" ", "_").toUpperCase());
-    }
-
+    errorBody.setErrorCode(""); // non-null to avoid NPE
     Matcher messageMatcher = HTML_ERROR_REGEX.matcher(body);
     if (messageMatcher.find()) {
-      errorBody.setMessage(messageMatcher.group(1).replaceAll("^[ .]+|[ .]+$", ""));
```
mihaimitrea-db, line 135: The tests seem to fail because this line was removed and not the if body is empty.

mihaimitrea-db (approved):
Otherwise LGTM

PR #760, opened 2026-04-15
Add support for authentication through Azure MSI

Divyansh-db:
## Summary

  Adds Azure Managed Service Identity (MSI) authentication to the Java SDK, porting the azure-msi credential strategy from the Go SDK (https://github.com/databricks/databricks-sdk-go/blob/main/config/auth_azure_msi.go).

##  Why

  The Java SDK had the azureUseMsi config field (ARM_USE_MSI) wired up but no provider behind it. Users on Azure VMs, App Service, or Functions with managed identities had to fall back to service principal secrets or Azure CLI. The Go SDK has supported this since early on — this PR closes the gap.

###  Interface changes

  - AzureMsiCredentialsProvider — New CredentialsProvider with auth type "azure-msi". Activated when isAzure() &&
  ARM_USE_MSI=true && (host or resourceId is set).
  - AzureMsiTokenSource — TokenSource that fetches tokens from the Azure IMDS endpoint
  (http://169.254.169.254/metadata/identity/oauth2/token). Supports system-assigned and user-assigned identities (via ARM_CLIENT_ID).

###  Behavioral changes

  - Default credential chain now includes azure-msi between azure-github-oidc and azure-client-secret, matching Go SDK ordering.
  - Produces Authorization, X-Databricks-Azure-SP-Management-Token, and X-Databricks-Azure-Workspace-Resource-Id headers.

###  Internal changes

  - MsiTokenResponse POJO for IMDS JSON deserialization (mirrors Go's msiToken struct), handling expires_on (epoch
  seconds).
  - Null-safe isAzureUseMsi() helper for the boxed Boolean config field.
  - Reuses AzureUtils.ensureHostPresent() for workspace URL resolution, consistent with other Azure providers.

 ## How is this tested?

  20 unit tests in AzureMsiCredentialsProviderTest across two groups:

  - TokenSourceTests (12) — IMDS request shape, response parsing, table-driven client_id handling (3 cases),
  table-driven invalid responses (5 cases: missing fields, bad expiry), HTTP error paths.
  - ProviderTests (8) — Mirrors Go SDK's TestMsiHappyFlow, TestMsiHappyFlowWithHostAndNoResourceID,
  TestMsiFailsOnResolveWorkspace, TestMsiTokenNotFound, plus table-driven null-return preconditions.
  - Verified it works against a real Azure VM


databricks-sdk-java/src/main/java/com/databricks/sdk/core/AzureMsiCredentialsProvider.java
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Null-safe check for the azureUseMsi config flag. The underlying field is a boxed Boolean, but
+   * the getter auto-unboxes to primitive boolean, which would NPE when the field is null. This
+   * helper treats null as false.
+   */
+  private static boolean isAzureUseMsi(DatabricksConfig config) {
+    try {
+      return config.getAzureUseMsi();
+    } catch (NullPointerException e) {
```
hectorcast-db: Causing a null-pointer on purpose and using a catch is very unorthodox. Better use:
```
Boolean flag = config.getAzureUseMsi();
return flag != null && flag;
```

hectorcast-db (approved):
Approved but change the issue above

PR #788, opened 2026-05-04
Canonicalize OAuth Bearer scheme when building Authorization header

mkazia:
## Summary
- Identity providers may return `token_type` in any case (e.g. `bearer`, `BEARER`) per RFC 6749/6750, but some downstream servers and proxies reject anything other than the canonical `Bearer`. This caused intermittent auth failures depending on the IdP's response casing.
- Adds `Token.getCanonicalTokenType()`, which returns `"Bearer"` whenever `tokenType` case-insensitively matches `"bearer"` and otherwise returns the original value untouched.
- Routes the three `Authorization` header construction sites through the new helper: `OAuthHeaderFactory.fromTokenSource`, `AzureCliCredentialsProvider`, and `ServingEndpointsDataPlaneImpl`. Non-Bearer schemes (e.g. `MAC`, custom) are unchanged.

## Test plan
- [x] `TokenTest.canonicalTokenTypeNormalizesBearerCasing` — covers `Bearer` / `bearer` / `BEARER` / `BeArEr` all canonicalize to `Bearer`
- [x] `TokenTest.canonicalTokenTypePreservesNonBearerSchemes` — `Custom` and `MAC` pass through unchanged
- [x] `OAuthHeaderFactoryTest` parameterized cases assert `Authorization: Bearer <token>` regardless of input casing; existing `Custom` case still passes (regression guard against over-canonicalization)
- [x] `mvn -pl databricks-sdk-java -am clean test -Dtest='TokenTest,OAuthHeaderFactoryTest'` — 12 passed

This pull request and its description were written by Isaac.


databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java
```diff
@@ -51,6 +51,22 @@ public String getTokenType() {
...
+   * unchanged.
+   *
+   * @return the canonicalized token type
+   */
+  public String getCanonicalTokenType() {
+    if ("bearer".equalsIgnoreCase(tokenType)) {
+      return "Bearer";
+    }
+    return tokenType;
+  }
```
renaudhartert-db, line 68: Let's avoid returning `null`. Could you update the method to either return an empty `String` or an empty `Optional<String>` if `tokenType` is `null`?

renaudhartert-db:
Looks good overall, with a small comment. Could you also make sure the `NEXT_CHANGELOG.md` file is updated with your change?

databricks-sdk-java/src/main/java/com/databricks/sdk/core/oauth/Token.java
renaudhartert-db: Never mind, I missed that we have a constructor check 👍

mkazia:
> Looks good overall, with a small comment. Could you also make sure the `NEXT_CHANGELOG.md` file is updated with your change?

done

PR #819, opened 2026-06-03
Fix Paginator dropping results on empty pages (DECO-27280)

renaudhartert-db:
The Paginator stopped iterating on the first page that returned zero items, even when that page carried a non-empty next_page_token. As a result, token-paginated calls such as tables().list(...) could silently return no results when the leading pages were empty but the real rows lived on later pages. This was reported in DECO-27280, where the Python SDK walked 99 pages (97 of them empty but with valid tokens) and returned 23 tables, while the Java SDK gave up on the first empty page and returned nothing.

flipNextPage now continues past empty pages and terminates only when nextPageFn returns null (or the response is null). This establishes a single contract for all pagination styles: nextPageFn returns null exactly when there are no more pages.

Offset-based pagination needed a matching change. The SCIM IAM list APIs and the redash 1-based list APIs relied on the Paginator stopping as soon as a page was empty, and their nextPageFn never returned null. Under the new behavior they would loop forever, because the offset (or page number) does not advance on a zero-item page. Their lambdas now return null on an empty page so they terminate explicitly. These files are generated by the SDK codegen template; the template is fixed in databricks-eng/universe-dev#11878, so the next regeneration reproduces the same guards. The generated edits are included here so this change is correct and mergeable on its own, in either order relative to the codegen change.

Adds PaginatorTest, which had no coverage before. It exercises empty leading and interleaved pages, offset-style termination, per-page fetch counts (forward progress with no redundant fetches), the fact that the loop has no built-in cap and depends entirely on nextPageFn returning null, error propagation through the loop, and the null-initial-request path used by legacy empty-request APIs.

This pull request and its description were written by Isaac.

### DIFF
--- a/databricks-sdk-java/src/main/java/com/databricks/sdk/service/iam/AccountGroupsAPI.java
+++ b/databricks-sdk-java/src/main/java/com/databricks/sdk/service/iam/AccountGroupsAPI.java
@@ -71,6 +71,9 @@ public class AccountGroupsAPI {
             impl::list,
             ListGroupsResponse::getResources,
             response -> {
+              if (response.getResources() == null || response.getResources().isEmpty()) {
+                return null;
+              }
               Long offset = request.getStartIndex();
               if (offset == null) {
                 offset = 0L;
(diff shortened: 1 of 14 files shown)


renaudhartert-db:
Closing as duplicate: https://github.com/databricks/databricks-sdk-java/pull/818

PR #900, opened 2026-08-26
Add WIF group assumption integration coverage

mihaimitrea-db:
## Summary

Adds an end-to-end integration test for workload identity federation group assumption.

The test creates a service principal and group, grants the service principal permission to assume the group, and verifies that:

- A WIF client using group_id can read a notebook granted only to the group.
- The same service principal without group assumption remains denied.

This is PR 3 of 3 and is stacked on #899.

## Testing

- Integration test sources compile
- Spotless and repository push checks

The integration test requires the existing CI environment and credentials to run.

NO_CHANGELOG=true

### DIFF
--- a/databricks-sdk-java/src/test/java/com/databricks/sdk/integration/DatabricksWifIT.java
+++ b/databricks-sdk-java/src/test/java/com/databricks/sdk/integration/DatabricksWifIT.java
@@ -1,15 +1,27 @@
 package com.databricks.sdk.integration;
 
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
 import com.databricks.sdk.AccountClient;
 import com.databricks.sdk.WorkspaceClient;
 import com.databricks.sdk.core.DatabricksConfig;
+import com.databricks.sdk.core.DatabricksException;
 import com.databricks.sdk.integration.framework.EnvContext;
 import com.databricks.sdk.integration.framework.EnvOrSkip;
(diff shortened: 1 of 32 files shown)


databricks-sdk-java/src/test/java/com/databricks/sdk/integration/DatabricksWifIT.java
```diff
@@ -18,6 +30,196 @@
...
+                      .delete(
+                          new DeleteWorkspaceAssignmentRequest()
+                              .setWorkspaceId(workspaceId)
+                              .setPrincipalId(servicePrincipalId)))) {
+        // Create the group that represents the temporary workspace role.
+        AccountGroup group =
+            a.groupsV2()
+                .create(
+                    new CreateAccountGroupRequest()
+                        .setDisplayName(NameUtils.uniqueName("java-sdk-wif-role-group")));
```
Divyansh-db, line 89: can we make these names suffixed with random characters so that concurrent runs of integration tests do not affect each other?

mihaimitrea-db: The `uniqueName` function already does that:

```java
  public static String uniqueName(String prefix) {
    return prefix + "-" + generateRandomBase16String(16);
  }
```

Divyansh-db (approved):

